PR #29, opened 2024-05-16
Spike extend compile to handle WhereNode.

WaVEV:
In this PR, I implemented a quick and dirty version of a local compiler. Note that I called it `_compile` to avoid overwriting the parent method (it should be overwritten, but to make a point, I decided not to).

As we can see, the idea is to take advantage of the Django objects and manage them as a tree. This approach would be easier and more understandable.

Note that each Node has a function `as_mql`, but we cannot add this method to the Node class. Therefore, I suggest using it as an abstract datatype. (It is a bit unconventional, given that we are using OOD, but it is better than the alternative.)

This implementation passes the same tests as the original one in the basic suite (it is small, but just to prove a concept, I think it is okay).

### DIFF
--- a/django_mongodb/compiler.py
+++ b/django_mongodb/compiler.py
@@ -1,5 +1,8 @@
+import re
+
+from bson.objectid import ObjectId
 from django.conf import settings
-from django.core.exceptions import EmptyResultSet
+from django.core.exceptions import EmptyResultSet, FullResultSet
 from django.db import (
     DatabaseError,
     IntegrityError,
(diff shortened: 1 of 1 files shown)


timgraham:
Instead of a long list of `isinstance` checks, I think we can patch `as_mql` onto each lookup, etc. as in https://github.com/Snowflake-Labs/django-snowflake/blob/e2962b369e16a38ef718008282578af003d06684/django_snowflake/lookups.py.

> Note that I called it _compile to avoid overwriting the parent method (it should be overwritten, but to make a point, I decided not to).

Overriding `compile()` would break the ability to use MongoDB alongside a SQL database.

WaVEV:
I thought we couldn't patch it (the nodes will not have the method even if
we patched it). So perfect, yes we can do it.

WaVEV:
> Instead of a long list of `isinstance` checks, I think we can patch `as_mql` onto each lookup, etc. as in https://github.com/Snowflake-Labs/django-snowflake/blob/e2962b369e16a38ef718008282578af003d06684/django_snowflake/lookups.py.
> 
> > Note that I called it _compile to avoid overwriting the parent method (it should be overwritten, but to make a point, I decided not to).
> 
> Overriding `compile()` would break the ability to use MongoDB alongside a SQL database.



> Instead of a long list of `isinstance` checks, I think we can patch `as_mql` onto each lookup, etc. as in https://github.com/Snowflake-Labs/django-snowflake/blob/e2962b369e16a38ef718008282578af003d06684/django_snowflake/lookups.py.
> 
> > Note that I called it _compile to avoid overwriting the parent method (it should be overwritten, but to make a point, I decided not to).
> 
> Overriding `compile()` would break the ability to use MongoDB alongside a SQL database.

Any idea where will we put all these stuff? or just create a large file with all lookups?

timgraham:
I imagine at least a few files: lookups.py, functions.py, nodes.py.

timgraham:
Superseded by https://github.com/mongodb-labs/django-mongodb/pull/33.

PR #94, opened 2024-08-04
add support for ordering by expressions

WaVEV:
The compiler can now handle aggregated fields in ordering, and several crashes have been fixed.

### DIFF
--- a/django_mongodb/compiler.py
+++ b/django_mongodb/compiler.py
@@ -6,9 +6,10 @@ from django.core.exceptions import EmptyResultSet, FullResultSet
 from django.db import DatabaseError, IntegrityError, NotSupportedError
 from django.db.models import Count, Expression
 from django.db.models.aggregates import Aggregate, Variance
-from django.db.models.expressions import Col, Ref, Value
+from django.db.models.expressions import Case, Col, Ref, Value, When
 from django.db.models.functions.comparison import Coalesce
 from django.db.models.functions.math import Power
+from django.db.models.lookups import IsNull
 from django.db.models.sql import compiler
 from django.db.models.sql.constants import GET_ITERATOR_CHUNK_SIZE, MULTI, SINGLE
 from django.utils.functional import cached_property
@@ -349,8 +350,14 @@ class SQLCompiler(compiler.SQLCompiler):
         ordering_fields, sort_ordering, extra_fields = self._get_ordering()
         query.project_fields = self.get_project_fields(columns, ordering_fields)
         query.ordering = sort_ordering
-        if extra_fields and columns is None:
-            query.extra_fields = self.get_project_fields(extra_fields)
+        # If columns is None, then get_project_fields() won't add
+        # ordering_fields to $project. Use $addFields (extra_fields) instead.
+        if columns is None:
+            extra_fields += ordering_fields
+        if extra_fields:
+            query.extra_fields = {
+                field_name: expr.as_mql(self, self.connection) for field_name, expr in extra_fields
+            }
         where = self.get_where()
         try:
             expr = where.as_mql(self, self.connection) if where else {}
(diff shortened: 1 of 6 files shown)


django_mongodb/compiler.py
```diff
@@ -473,12 +451,47 @@ def get_project_fields(self, columns=None, ordering=None):
...
+        - A tuple of ('field_name': Expression, ...) for expressions that need
+          to be added to extra_fields.
+        """
+        fields = {}
+        sort_ordering = SON()
+        extra_fields = {}
+        idx = itertools.count(start=1)
+        for order in self.order_by_expressions or []:
+            if isinstance(order.expression, Ref):
+                # TODO change?
```
timgraham, line 474: todo: understand this to know whether it makes a difference if we use the commented version or the uncommented version.

WaVEV: There are no differences.  is okey to just use refs, but maybe we broke an abstraction? but in all part of the code we do (if this is the case)

django_mongodb/compiler.py
```diff
@@ -473,12 +451,47 @@ def get_project_fields(self, columns=None, ordering=None):
...
+                # field_name = order.expression.as_mql(self, self.connection).removeprefix("$")
+                # extra_fields[field_name] = order.expression
+                field_name = order.expression.refs
+            elif isinstance(order.expression, Col):
+                field_name = order.expression.as_mql(self, self.connection).removeprefix("$")
+                fields[field_name] = order.expression
+            else:
+                # The expression must be added to extra_fields with an alias.
+                field_name = f"__order{next(idx)}"
+                fields[field_name] = order.expression
```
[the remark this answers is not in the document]
WaVEV: the reason is:
The `refs` are actually references to specific fields selected in a query. These references compile to a position number within the `as_sql` method. To address this, I decided to move the `refs` to the post-project fields. Additionally, some of these `refs` are calculated during the project stage (such as annotations), so the `refs` are processed after the project stage. But the other projected fields that aren't `refs` have all the info to make the full expression.

Btw, 'refs' also has all the info to be compiled well, like an annotation with name (something that is not  `__annotation` but isn't a column), it has the reference but also the full expression with it. So if we want to delete the extra field we could, the downside will be a field will be calculated twice or more.

django_mongodb/query.py
```diff
@@ -51,28 +50,11 @@ def __init__(self, compiler):
         self.lookup_pipeline = None
         self.project_fields = None
         self.aggregation_pipeline = compiler.aggregation_pipeline
+        self.extra_fields = None
```
WaVEV, line 53: I think now we can pass the compiler alone and the query will know how to fill in all the fields from the compiler.

django_mongodb/compiler.py
WaVEV: 🤔 mmh now I see the idea, and you are right, we could ignore the refs projection, we don't need them, because they are already managed in $project. So we can remove the refs from the extra fields

django_mongodb/aggregates.py
```diff
@@ -47,7 +47,7 @@ def count(self, compiler, connection, resolve_inner_expression=False, **extra_co
             source_expressions = node.get_source_expressions()
             filter_ = deepcopy(self.filter)
             filter_.add(
-                WhereNode([Exact(source_expressions[0], Value(None))], negated=True),
+                WhereNode([IsNull(source_expressions[0], True)], negated=True),
```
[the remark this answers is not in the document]
WaVEV: Yes, it doesn't affect any current test, but I don't want to forget this tiny change. This IsNull is needed when the source_expression[0] is a foreign field. maybe we should create another ticket in order to make it more traceable ?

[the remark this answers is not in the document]
timgraham: Ok, thanks. We could merge it now. Just put it in a separate PR with the explanation, please. I'll remove it here when I squash the final edits.

django_mongodb/compiler.py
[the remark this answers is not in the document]
timgraham: Even with the explanation, it's still difficult to understand. Now in this method, `extra_fields` is only used for nulls first/last, but then outside this method, all `ordering_fields` are added to `extra_fields` when `columns is None` (I guess because they won't be added to `$project` in that case).

django_mongodb/query.py
timgraham: This is an idea for a later cleanup, right?

timgraham (approved):

django_mongodb/query.py
WaVEV: yes, to avoid doing 
```python
query.something = value
query.anotherthing = other_value
...
```

PR #111, opened 2024-08-20
POC support update with expressions.

WaVEV:
This is a quick approach to support expression in updates. I have to say that the update pipeline is not expressive as it's needed


django_mongodb/compiler.py
```diff
@@ -591,42 +603,44 @@ def execute_sql(self, result_type):
...
+        try:
+            criteria = self.build_query().mongo_query
+        except EmptyResultSet:
+            return 0
         is_empty = not bool(values)
-        rows = 0 if is_empty else self.update(values)
+        if is_empty:
+            rows = 0
+        else:
+            base_pipeline = [
```
WaVEV, line 617: This part is the only thing that need focus. Here I just create the **$merge** pipeline and the **affected rows** pipeline with a single transaction. I really don't know if this is the way to go.

Jibola: This is the right way to go about it. having a transaction ensures that state isn't changed between the lookup and the update.

django_mongodb/compiler.py
```diff
@@ -591,42 +603,44 @@ def execute_sql(self, result_type):
...
+                    "$merge": {
+                        "into": self.collection_name,
+                        "whenMatched": "replace",
+                        "whenNotMatched": "discard",
+                    }
+                },
+            ]
+            with self.connection.connection.start_session() as session, session.start_transaction():
+                result = next(self.collection.aggregate(count_pipeline), {"count": 0})
+                self.collection.aggregate(pipeline)
```
Jibola, line 634: what has been the output when you merge:
`[*base_pipeline, {$merge..}, {$count...}]` Does this not work whatsoever?

Jibola (commented):
Overall, agree with the approach. I'm *hoping* there is another way to combine the count and merge _after_ the merge has occurred, since we ultimately just want the number of documents that were changed.

django_mongodb/compiler.py
WaVEV: I think we can use update many directly as Shane mentioned. If I've re-read the docs  and what the docs support are this three stages.

WaVEV: Yeap, It works 🚀. So we are able to do the update without making two queries (one for count and the other for the update)

WaVEV:
The final implementation is in #112

PR #130, opened 2024-09-17
fix incorrect GenericRelation joining

WaVEV:
By adding support for Field.get_extra_restriction().

### DIFF
--- a/django_mongodb/query.py
+++ b/django_mongodb/query.py
@@ -3,7 +3,7 @@ from operator import add as add_operator
 
 from django.core.exceptions import EmptyResultSet, FullResultSet
 from django.db import DatabaseError, IntegrityError, NotSupportedError
-from django.db.models.expressions import Case, When
+from django.db.models.expressions import Case, Col, When
 from django.db.models.functions import Mod
 from django.db.models.lookups import Exact
 from django.db.models.sql.constants import INNER
@@ -105,6 +105,7 @@ def join(self, compiler, connection):
     lhs_fields = []
     rhs_fields = []
     # Add a join condition for each pair of joining fields.
+    parent_template = "parent__field__"
     for lhs, rhs in self.join_fields:
         lhs, rhs = connection.ops.prepare_join_on_clause(
             self.parent_alias, lhs, compiler.collection_name, rhs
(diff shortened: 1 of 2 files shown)


.github/workflows/test-python.yml
```diff
@@ -43,8 +43,8 @@ jobs:
       - name: Checkout Django
         uses: actions/checkout@v4
         with:
-          repository: 'mongodb-forks/django'
-          ref: 'mongodb-5.0.x'
+          repository: 'wavev/django'
+          ref: 'add-support-join-extra-conditions'
```
Jibola, line 47: I imagine this should be reverted

django_mongodb/fields/object_id.py
```diff
@@ -0,0 +1,30 @@
+from bson import ObjectId, errors
```
Jibola, line 1: is this file a required part of the change?

django_mongodb/query.py
```diff
@@ -105,8 +106,28 @@ def join(self, compiler, connection):
         # In the lookup stage, the reference to this column doesn't include
         # the collection name.
         rhs_fields.append(rhs.as_mql(compiler, connection).replace(f"{self.table_name}.", "", 1))
+    extra = self.join_field.get_extra_restriction(self.table_alias, self.parent_alias)
+    if extra:
+        columns = []
+        for expr in extra.leaves():
+            if hasattr(expr, "lhs") and isinstance(expr.lhs, Col):
+                columns.append((expr.lhs, len(lhs_fields)))
```
Jibola, line 114: To confirm, we use len(lhs_fields) because expr.lhs will now be the newest column index, thus equal to len(lhs_fields)?

django_mongodb/query.py
```diff
@@ -105,8 +106,28 @@ def join(self, compiler, connection):
         # In the lookup stage, the reference to this column doesn't include
         # the collection name.
         rhs_fields.append(rhs.as_mql(compiler, connection).replace(f"{self.table_name}.", "", 1))
+    extra = self.join_field.get_extra_restriction(self.table_alias, self.parent_alias)
+    if extra:
```
Jibola, line 110: Could we add a code comments explaining the purpose of the `extra` section?

django_mongodb/query.py
```diff
@@ -105,8 +106,28 @@ def join(self, compiler, connection):
...
+        columns = []
+        for expr in extra.leaves():
+            if hasattr(expr, "lhs") and isinstance(expr.lhs, Col):
+                columns.append((expr.lhs, len(lhs_fields)))
+                lhs_fields.append(expr.lhs.as_mql(compiler, connection))
+            if hasattr(expr, "rhs") and isinstance(expr.rhs, Col):
+                columns.append((expr.rhs, None))
```
Jibola, line 117: Why does rhs not get the same as lhs. Can it not be referenced by number still?

django_mongodb/query.py
```diff
@@ -105,8 +106,28 @@ def join(self, compiler, connection):
...
+        replacements = {}
+        for col, parent_pos in columns:
+            column_target = col.copy()
+            if column_target.alias == self.table_name:
+                column_target.alias = compiler.collection_name
+            else:
+                column_target.target.db_column = f"{parent_template}{parent_pos}"
+                column_target.target.set_attributes_from_name(f"{parent_template}{len(lhs_fields)}")
```
Jibola, line 125: why do we need to use len(lhs_fields) instead of `parent_pos`

Jibola (changes requested):
Left commentary to get clarification on things

.github/workflows/test-python.yml
WaVEV: It will.

django_mongodb/fields/object_id.py
WaVEV: This new field was used in the django library when a new model is created, not here. But we will handle those fields as string for now, so this field is not longer needed.

django_mongodb/query.py
WaVEV: You are right, it should be the same.

django_mongodb/query.py
WaVEV: Fixed

django_mongodb/query.py
WaVEV: Yes, it is.

timgraham (approved):

PR #176, opened 2024-10-26
remove unnecessary $convert in ObjectIdAutoField $lookup queries

WaVEV:


timgraham (approved):
I added a regression test, okay?

WaVEV:
> I added a regression test, okay?

Yes, it is.

PR #195, opened 2024-11-26
INTPYTHON-424 add django_mongodb.parse_uri() to configure DATABASES

aclark4life:
Supersedes #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,16 +110,58 @@ to this:
 DATABASES = {
     "default": {
         "ENGINE": "django_mongodb",
+        "HOST": "mongodb+srv://cluster0.example.mongodb.net",
         "NAME": "my_database",
         "USER": "my_user",
         "PASSWORD": "my_password",
-        "OPTIONS": {...},
+        "PORT": 27017,
+        "OPTIONS": {
+            # Example:
+            "retryWrites": "true",
+            "w": "majority",
+            "tls": "false",
+        },
     },
 }
 ```
 
+For a localhost configuration, you can omit `HOST` or specify
+`"HOST": "localhost"`.
+
+`HOST` only needs a scheme prefix for SRV connections (`mongodb+srv://`). A
+`mongodb://` prefix is never required.
+
 `OPTIONS` is an optional dictionary of parameters that will be passed to
 [`MongoClient`](https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html).
+
+`USER`, `PASSWORD`, and `PORT` (if 27017) may also be optional.
+
+For a replica set or sharded cluster where you have multiple hosts, include
+all of them in `HOST`, e.g.
+`"mongodb://mongos0.example.com:27017,mongos1.example.com:27017"`.
+
+Alternatively, if you prefer to simply paste in a MongoDB URI rather than parse
+it into the format above, you can use:
+
+```python
+import django_mongodb
+
+MONGODB_URI = "mongodb+srv://my_user:my_password@cluster0.example.mongodb.net/myDatabase?retryWrites=true&w=majority&tls=false"
+DATABASES["default"] = django_mongodb.parse_uri(MONGODB_URI)
+```
+
+This constructs a `DATABASES` setting equivalent to the first example.
+
+#### `django_mongodb.parse_uri(uri, conn_max_age=0, test=None)`
+
+`parse_uri()` provides a few options to customize the resulting `DATABASES`
+setting, but for maximum flexibility, construct `DATABASES` manually as
+described above.
+
+- Use `conn_max_age` to configure [persistent database connections](
+  https://docs.djangoproject.com/en/stable/ref/databases/#persistent-database-connections).
+- Use `test` to provide a dictionary of [settings for test databases](
+  https://docs.djangoproject.com/en/stable/ref/settings/#test).
 
 Congratulations, your project is ready to go!
 
--- a/django_mongodb/__init__.py
+++ b/django_mongodb/__init__.py
@@ -2,7 +2,7 @@ __version__ = "5.0a0"
 
 # Check Django compatibility before other imports which may fail if the
 # wrong version of Django is installed.
-from .utils import check_django_compatability
+from .utils import check_django_compatability, parse_uri
 
 check_django_compatability()
 
@@ -13,6 +13,8 @@ from .functions import register_functions  # noqa: E402
 from .indexes import register_indexes  # noqa: E402
 from .lookups import register_lookups  # noqa: E402
 from .query import register_nodes  # noqa: E402
+
+__all__ = ["parse_uri"]
 
 register_aggregates()
 register_expressions()
--- a/django_mongodb/utils.py
+++ b/django_mongodb/utils.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db.backends.utils import logger
 from django.utils.version import get_version_tuple
+from pymongo.uri_parser import parse_uri as pymongo_parse_uri
 
 
 def check_django_compatability():
@@ -23,6 +24,38 @@ def check_django_compatability():
             f"You must use the latest version of django-mongodb {A}.{B}.x "
             f"with Django {A}.{B}.y (found django-mongodb {__version__})."
         )
+
+
+def parse_uri(uri, conn_max_age=0, test=None):
+    """
+    Convert the given uri into a dictionary suitable for Django's DATABASES
+    setting.
+    """
+    uri = pymongo_parse_uri(uri)
+    host = None
+    port = None
+    if uri["fqdn"]:
+        # This is a SRV URI and the host is the fqdn.
+        host = f"mongodb+srv://{uri['fqdn']}"
+    else:
+        nodelist = uri.get("nodelist")
+        if len(nodelist) == 1:
+            host, port = nodelist[0]
+        elif len(nodelist) > 1:
+            host = ",".join([f"{host}:{port}" for host, port in nodelist])
+    settings_dict = {
+        "ENGINE": "django_mongodb",
+        "NAME": uri["database"],
+        "HOST": host,
+        "PORT": port,
+        "USER": uri.get("username"),
+        "PASSWORD": uri.get("password"),
+        "OPTIONS": uri.get("options"),
+        "CONN_MAX_AGE": conn_max_age,
+    }
+    if test:
+        settings_dict["TEST"] = test
+    return settings_dict
 
 
 def set_wrapped_methods(cls):
--- a/tests/backend_/utils/test_parse_uri.py
+++ b/tests/backend_/utils/test_parse_uri.py
@@ -1,0 +1,71 @@
+from unittest.mock import patch
+
+import pymongo
+from django.test import SimpleTestCase
+
+from django_mongodb import parse_uri
+
+
+class ParseURITests(SimpleTestCase):
+    def test_simple_uri(self):
+        settings_dict = parse_uri("mongodb://cluster0.example.mongodb.net/myDatabase")
+        self.assertEqual(settings_dict["ENGINE"], "django_mongodb")
+        self.assertEqual(settings_dict["NAME"], "myDatabase")
+        self.assertEqual(settings_dict["HOST"], "cluster0.example.mongodb.net")
+
+    def test_no_database(self):
+        settings_dict = parse_uri("mongodb://cluster0.example.mongodb.net")
+        self.assertIsNone(settings_dict["NAME"])
+        self.assertEqual(settings_dict["HOST"], "cluster0.example.mongodb.net")
+
+    def test_srv_uri_with_options(self):
+        uri = "mongodb+srv://my_user:my_password@cluster0.example.mongodb.net/my_database?retryWrites=true&w=majority"
+        # patch() prevents a crash when PyMongo attempts to resolve the
+        # nonexistent SRV record.
+        with patch("dns.resolver.resolve"):
+            settings_dict = parse_uri(uri)
+        self.assertEqual(settings_dict["NAME"], "my_database")
+        self.assertEqual(settings_dict["HOST"], "mongodb+srv://cluster0.example.mongodb.net")
+        self.assertEqual(settings_dict["USER"], "my_user")
+        self.assertEqual(settings_dict["PASSWORD"], "my_password")
+        self.assertIsNone(settings_dict["PORT"])
+        self.assertEqual(
+            settings_dict["OPTIONS"], {"retryWrites": True, "w": "majority", "tls": True}
+        )
+
+    def test_localhost(self):
+        settings_dict = parse_uri("mongodb://localhost")
+        self.assertEqual(settings_dict["HOST"], "localhost")
+        self.assertEqual(settings_dict["PORT"], 27017)
+
+    def test_localhost_with_port(self):
+        settings_dict = parse_uri("mongodb://localhost:27018")
+        self.assertEqual(settings_dict["HOST"], "localhost")
+        self.assertEqual(settings_dict["PORT"], 27018)
+
+    def test_hosts_with_ports(self):
+        settings_dict = parse_uri("mongodb://localhost:27017,localhost:27018")
+        self.assertEqual(settings_dict["HOST"], "localhost:27017,localhost:27018")
+        self.assertEqual(settings_dict["PORT"], None)
+
+    def test_hosts_without_ports(self):
+        settings_dict = parse_uri("mongodb://host1.net,host2.net")
+        self.assertEqual(settings_dict["HOST"], "host1.net:27017,host2.net:27017")
+        self.assertEqual(settings_dict["PORT"], None)
+
+    def test_conn_max_age(self):
+        settings_dict = parse_uri("mongodb://localhost", conn_max_age=600)
+        self.assertEqual(settings_dict["CONN_MAX_AGE"], 600)
+
+    def test_test_kwarg(self):
+        settings_dict = parse_uri("mongodb://localhost", test={"NAME": "test_db"})
+        self.assertEqual(settings_dict["TEST"], {"NAME": "test_db"})
+
+    def test_invalid_credentials(self):
+        msg = "The empty string is not valid username."
+        with self.assertRaisesMessage(pymongo.errors.InvalidURI, msg):
+            parse_uri("mongodb://:@localhost")
+
+    def test_no_scheme(self):
+        with self.assertRaisesMessage(pymongo.errors.InvalidURI, "Invalid URI scheme"):
+            parse_uri("cluster0.example.mongodb.net")


aclark4life:
This is working for me with an Atlas URI

Screenshot 2024-11-25 at 9 01 50 PM

django_mongodb/utils.py
```diff
@@ -25,6 +26,39 @@ def check_django_compatability():
...
+                isinstance(first_node, tuple)
+                and len(first_node) > 1
+                and isinstance(first_node[1], int)
+            ):
+                port = first_node[1]
+    else:
+        # The fqdn is not none so we need to add the mongodb+srv:// prefix to the host.
+        host = f"mongodb+srv://{uri['fqdn']}"
+
+    url = {
```
timgraham, line 48: url -> settings_dict (to match DatabaseWrapper attribut of the same name).

django_mongodb/utils.py
```diff
@@ -25,6 +26,39 @@ def check_django_compatability():
         )
 
 
+def parse(uri):
+    uri = parse_uri(uri)
+    port = None
+
+    if uri["fqdn"] is None:
+        # If fqdn is None then this is not a SRV URI, so we need to extract the
```
timgraham, line 34: Personal preference: write comments without "we".   "so extra..." saves a few words.

django_mongodb/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 # Check Django compatibility before other imports which may fail if the
 # wrong version of Django is installed.
-from .utils import check_django_compatability
+from .utils import check_django_compatability, parse
```
timgraham, line 5: This file should have an `__all__ = ["parse"]` to clarify the public API.

django_mongodb/utils.py
```diff
@@ -25,6 +26,39 @@ def check_django_compatability():
...
+            ):
+                port = first_node[1]
+    else:
+        # The fqdn is not none so we need to add the mongodb+srv:// prefix to the host.
+        host = f"mongodb+srv://{uri['fqdn']}"
+
+    settings_dict = {
+        "ENGINE": "django_mongodb",
+        "NAME": uri["database"],
+        "HOST": host,
```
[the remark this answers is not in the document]
timgraham: It seems to me that kind of goes against the point of parsing the URI in the first place. The expected behavior is unclear if there is some bug in the parsing that causes a mismatch is parsed vs. unparsed (like at https://github.com/mongodb-labs/django-mongodb/blob/36c57187de430ab4313e8db5c3ce8ab2ab569f5b/django_mongodb/base.py#L162-L166 what takes precedence, the values in HOST or the other values). And then if the dbname remains in the URI, what of code like https://github.com/mongodb-labs/django-mongodb/blob/36c57187de430ab4313e8db5c3ce8ab2ab569f5b/django_mongodb/creation.py#L13-L14 that overrides it?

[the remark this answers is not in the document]
aclark4life: We cannot pass the raw URI as host because we need to support the Django SETTINGS dictionary API.

django_mongodb/utils.py
```diff
@@ -25,6 +26,39 @@ def check_django_compatability():
...
+        "USER": uri.get("username"),
+        "PASSWORD": uri.get("password"),
+    }
```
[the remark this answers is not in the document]
aclark4life: I'm not sure what the right thing to do here is but maybe something like https://github.com/jazzband/dj-database-url/blob/358746c41d022be2bab6dc7fa0ac1b98d59069f2/dj_database_url/__init__.py#L176-L189

django_mongodb/utils.py
```diff
@@ -25,6 +26,39 @@ def check_django_compatability():
         )
 
 
+def parse(uri):
```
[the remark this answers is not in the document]
aclark4life: Maybe! I'm relying on https://github.com/jazzband/dj-database-url/blob/358746c41d022be2bab6dc7fa0ac1b98d59069f2/dj_database_url/__init__.py#L95-L110 for inspiration.

django_mongodb/utils.py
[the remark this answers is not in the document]
aclark4life: Done in fb13cbb2e0adeeebf16f4515933451c66fc5c342

django_mongodb/utils.py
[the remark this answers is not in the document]
aclark4life: Done in 4ba50ff3540eace436ed71acb531750d1495be90

README.md
```diff
@@ -121,6 +121,13 @@ DATABASES = {
 `OPTIONS` is an optional dictionary of parameters that will be passed to
 [`MongoClient`](https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html).
 
+Alternatively, you can use a MongoDB URI with `django_mongodb.parse(uri)`:
```
Jibola, line 124: Let's have this be the standard rather than the alternative.

django_mongodb/utils.py
```diff
@@ -25,6 +26,37 @@ def check_django_compatability():
...
+    port = None
+
+    if uri["fqdn"] is not None:
+        # If the fqdn is present, this is a SRV URI and the host is the fqdn.
+        host = f"mongodb+srv://{uri['fqdn']}"
+    else:
+        # If the fqdn is not present, this is a standard URI and the host and
+        # port are in the nodelist.
+        nodelist = [f"{node[0]}:{node[1]}" for node in uri["nodelist"]]
+        host, port = nodelist[0].split(":")
```
Jibola, line 42: I was thinking we could do this (since this can be passed to the `MongoClient`)

```suggestion
        host = [f"{node[0]}:{node[1]}" for node in uri["nodelist"]]
        _, port = host[0].split(":")
```

django_mongodb/utils.py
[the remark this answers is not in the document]
timgraham: I'm not sure `**kwargs` is the best API design. It allows passing in additional information not in the URI, but it also allows providing information that would override what's in the URI. Like maybe I pass in `OPTIONS={'foo': 'bar'}}` and I don't realize this silently overwrites the options that are in the URI! The Zen of Python says, "There should be one-- and preferably only one --obvious way to do it."  If the user wants to provide a URI, I think they should be required put all the information the URI accepts in there. If some information needs to be provided to DATABASES that doesn't fit in the URI, then we could add a way to do that.

Jibola (changes requested):

django_mongodb/utils.py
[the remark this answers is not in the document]
aclark4life: Right, `dj-database-url` does this:

```

def parse(
    url: str,
    engine: Optional[str] = None,
    conn_max_age: Optional[int] = 0,
    conn_health_checks: bool = False,
    disable_server_side_cursors: bool = False,
    ssl_require: bool = False,
    test_options: Optional[Dict[str, Any]] = None,
) -> DBConfig:
```

README.md
aclark4life: Done in 0484032244d8f905cdc12c904ed6c308b67fe6b5!

django_mongodb/utils.py
aclark4life: Actually, if `nodelist > 1` I think we can just pass `nodelist`. Otherwise if `nodelist == 1` then maybe we set host and port. Port is only getting passed outside of the `PORT` setting when replica set.

aclark4life: Fixed in 58cdeb0a957b200253712d44992a88c77f597cf7, 🤞

aclark4life:
Via @timgraham we may look to https://github.com/jazzband/dj-database-url/blob/master/tests/test_dj_database_url.py for some testing inspiration.

django_mongodb/utils.py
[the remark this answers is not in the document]
aclark4life: Back to no kwargs for now until we decide what specific kwargs to support

aclark4life:
@Jibola @timgraham I think this is ready except for:

- What specific kwargs to accept in `parse()`
- What additional tests should be added
- What have I missed if anything

Tests for this are running in Actions.

README.md
```diff
@@ -103,8 +103,15 @@ $ python manage.py startapp myapp --template https://github.com/mongodb-labs/dja
 
 ### Configuring the `DATABASES` setting
 
-After you've set up a project, configure Django's `DATABASES` setting similar
-to this:
+After you've set up a project, configure Django's `DATABASES` with
```
timgraham, line 106: Are you going to update the project template? This also needs `import django_mongodb`. Should I use a URI even with a localhost configuration?

README.md
```diff
@@ -103,8 +103,15 @@ $ python manage.py startapp myapp --template https://github.com/mongodb-labs/dja
...
+`django_mongodb.parse(uri)`:
+
+```python
+MONGODB_URI = "mongodb://<my_user>:<my_password>@localhost:27017/my_database"
+DATABASES["default"] = django_mongodb.parse(MONGODB_URI)
+```
+
+Alternatively, you can configure the DATABASES settings directly:
```
timgraham, line 114: Explain why would I want or need to use the alternative. Add backticks around DATABASES.

django_mongodb/utils.py
```diff
@@ -25,6 +26,33 @@ def check_django_compatability():
         )
 
 
+def parse(uri):
```
timgraham, line 29: Please add a docstring.

tests/utils_/test_parse.py
```diff
@@ -0,0 +1,26 @@
...
+        srv_record = MagicMock()
+        srv_record.target.to_text.return_value = "cluster0.example.mongodb.net"
+        mock_resolver.return_value = [srv_record]
+        settings_dict = django_mongodb.parse(MONGODB_URI)
+        self.assertEqual(settings_dict["ENGINE"], "django_mongodb")
+        self.assertEqual(settings_dict["NAME"], "myDatabase")
+        self.assertEqual(settings_dict["HOST"], "mongodb+srv://cluster0.example.mongodb.net")
+        self.assertEqual(settings_dict["USER"], "myDatabaseUser")
+        self.assertEqual(settings_dict["PASSWORD"], "D1fficultP@ssw0rd")
+        self.assertEqual(settings_dict["PORT"], None)
```
timgraham, line 26: OPTIONS

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.5.0
+  rev: v5.0.0
```
timgraham, line 4: Please move to another PR.

timgraham (commented):
Please test with various URIs to exercise the different code paths and the cases when values aren't present.

README.md
aclark4life: Yes I'll update the project template and add the import statement to the README, and yes to localhost, here's how I'm running the tests:

```
DATABASE_URL = os.environ.get("DATABASE_URL", "mongodb://localhost:27017/djangotests")
```

timgraham: In a similar spirit to my comment below, I think promoting a URL as the default (even for something like localhost) is sending people in a more complicated direction in the long run, especially for Django developers coming to MongoDB. It seems pretty simple to parse a URI to the DATABASES format just by looking at it, or as I suggested, use `print(parse_uri(...))` if in doubt.

aclark4life: Agree, I put it back the way it was, which fits the pattern of Django-way, followed by Mongo-way. @Jibola can you roll with this?

aclark4life: Looks like we don't need `DISABLE_SERVER_SIDE_CURSORS` because it is PG specific

Jibola: **TL;DR**: I concede to this and in terms of how we dictate instructions for the a quickstart or target audience, I think that can be broken out of this one. 

The two conventional ways of getting a local MongoDB URI are:
* through [local atlas](https://www.mongodb.com/docs/atlas/cli/current/atlas-cli-deploy-local/)
* through [`mongo-orchestration`](https://github.com/10gen/mongo-orchestration)
 but both of these methods -- in my opinion -- are still roads less traveled. 
 
I really do see the _average_ case being someone with an Atlas URI pointing to instructions on how to get an Atlas URI via the [Getting Started](https://www.mongodb.com/docs/atlas/getting-started/) page.

Because that average case ends up being a more involved cluster, I perceive the template shooting for those cases where no parsing or local db setup is done as the quickest way to get started. More advanced users in the Django and dev space will definitely be able to understand the parsing but true beginners to MongoDB I foresee still being confused.  

Nonetheless, ruminating over this has less to do with the implementation of `parse` and more to do with what gets spit out from our template. Since our template currently supports the `DATABASES` standard, we'll just keep the PR in alignment that way.

aclark4life:
> LGTM, pending a few more tests. @timgraham for the final ✅

OK added a couple more localhost tests.

Jibola:
Looks like there's still some linting issues.

README.md
```diff
@@ -121,6 +121,16 @@ DATABASES = {
 `OPTIONS` is an optional dictionary of parameters that will be passed to
 [`MongoClient`](https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html).
 
+Alternatively, you can configure Django's `DATABASES` with
```
timgraham, line 124: Still missing the motivation as to why the alternative is needed or should be chosen, as well as docs for the kwargs.

README.md
```diff
@@ -121,6 +121,16 @@ DATABASES = {
...
+`django_mongodb.parse(uri)`:
+
+```python
+import django_mongodb
+
+MONGODB_URI = "mongodb://<my_user>:<my_password>@localhost:27017/my_database"
```
timgraham, line 130: why use placeholders for user,pass,db, but not host/port?

django_mongodb/utils.py
```diff
@@ -25,6 +26,58 @@ def check_django_compatability():
...
+    settings. This function is a wrapper around PyMongo's
+    ``pymongo.uri_parser.parse_uri()`` function that converts PyMongo's
+    settings dictionary into a Django database settings dictionary.
+    """
+    uri = parse_uri(uri)
+
+    host = None
+    port = None
+
+    if not engine:
```
timgraham, line 43: As far as I can think, this would be useful if something is subclassing our backend. Seems unlikely. In any case,  you could simply make `engine="django_mongodb"` the default kwarg if you want to keep it.

timgraham (commented):
I thought I left this comment before but I can't find it. I think we should call the function something more than just `parse()`, like `parse_uri()`. Unlike `dj-database-url`, our library has many purposes and therefore many things that could be parsed.

aclark4life:
Thanks for the feedback, @Jibola and @timgraham ! Updates:
- Rename func
- Removed `engine` kwarg
- Added docs

django_mongodb/utils.py
```diff
@@ -25,6 +26,39 @@ def check_django_compatability():
...
+    port = None
+    if uri["fqdn"] is not None:
+        # If the fqdn is present, this is a SRV URI and the host is the fqdn.
+        host = f"mongodb+srv://{uri['fqdn']}"
+    else:
+        nodelist = uri.get("nodelist")
+        if len(nodelist) == 1:
+            host, port = nodelist[0]
+        elif len(nodelist) > 1:
+            host = ",".join([f"{host}:{port}" for host, port in nodelist])
```
timgraham, line 45: This branch appears to be untested (add an `assert False` and no tests fail). It looks like there may be a bug if port is None appending ":None" to the host.

Also, I see no tests for a non-None PORT.

aclark4life: Good catches, thanks! Fixed in 05df78c27580765c503d2a19eeb791ed271fbc9e and 05ba4453f8a2a71b5afbaf1828e57da9531e58ee

tests/utils_/test_parse_uri.py
```diff
@@ -0,0 +1,90 @@
...
+URI = "mongodb+srv://myDatabaseUser:D1fficultP%40ssw0rd@cluster0.example.mongodb.net/myDatabase?retryWrites=true&w=majority&tls=false"
+
+
+class ParseURITests(SimpleTestCase):
+    def setUp(self):
+        self.srv_record = MagicMock()
+        self.srv_record.target.to_text.return_value = "cluster0.example.mongodb.net"
+        self.patcher = patch("dns.resolver.resolve", return_value=[self.srv_record])
+        self.mock_resolver = self.patcher.start()
+        self.addCleanup(self.patcher.stop)
```
timgraham, line 17: Is the mocking needed? If I comment out the method (even with networking disabled) or change the return value to e.g. cluster1, I see no failures. Also, the [DNS resolving code](https://github.com/mongodb/mongo-python-driver/blob/1b89da4829945359ae99ecfb579104d2d28f7b27/pymongo/uri_parser.py#L557-L559) only seems applicable to srv URLs, but some tests with non-srv URIs are decorated with `@patch("dns.resolver.resolve")`. If any of this is needed, please add some explanatory comments.

tests/utils_/test_parse_uri.py
```diff
@@ -0,0 +1,90 @@
...
+        settings_dict = parse_uri("mongodb://cluster0.example.mongodb.net/myDatabase")
+        self.assertEqual(settings_dict["ENGINE"], "django_mongodb")
+        self.assertEqual(settings_dict["NAME"], "myDatabase")
+        self.assertEqual(settings_dict["HOST"], "cluster0.example.mongodb.net")
+
+    @patch("dns.resolver.resolve")
+    def test_no_database(self, mock_resolver):
+        settings_dict = parse_uri("mongodb://cluster0.example.mongodb.net/")
+        self.assertEqual(settings_dict["ENGINE"], "django_mongodb")
+        self.assertIsNone(settings_dict["NAME"])
```
Jibola, line 30: Turns out, based on the spec for `NAME` this should end up returning an empty string if it comes back as `None`.
https://docs.djangoproject.com/en/5.1/ref/settings/#name

Jibola: Additionally, in this case, I think it's best to still have the `name` as a passed through kwarg because Atlas URIs do often not provide a database name. I'll provide an example in the code above about how this would work.

django_mongodb/utils.py
```diff
@@ -25,6 +26,39 @@ def check_django_compatability():
...
+    settings_dict = {
+        "ENGINE": "django_mongodb",
+        "NAME": uri["database"],
```
Jibola, line 48:
```suggestion
        "NAME": name or uri["database"] or '',
```

Jibola (dismissed):

tests/utils_/test_parse_uri.py
aclark4life: If the URI does not contain a database name, then we don't need to add `NAME`. 1519e0f484682f2ad7b89bbb670c0c9c9fac0498

django_mongodb/utils.py
aclark4life: I would say let's just not add the key 1519e0f484682f2ad7b89bbb670c0c9c9fac0498

django_mongodb/utils.py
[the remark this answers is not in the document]
aclark4life: This works as-is because `port` is always in the results of PyMongo's `parse_uri` …

tests/utils_/test_parse_uri.py
timgraham: I don't think the None/'' distinction matters. What does it mean to use Django + MongoDB without specifying a database name? I think NAME must be required, although it's not the job of `parse_uri()` to raise an error. Currently, when trying runserver without NAME:
```
  File "/opt/python3.12.0/lib/python3.12/threading.py", line 1052, in _bootstrap_inner
    self.run()
  File "/opt/python3.12.0/lib/python3.12/threading.py", line 989, in run
    self._target(*self._args, **self._kwargs)
  File "/home/tim/code/django/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/home/tim/code/django/django/core/management/commands/runserver.py", line 136, in inner_run
    self.check_migrations()
  File "/home/tim/code/django/django/core/management/base.py", line 581, in check_migrations
    executor = MigrationExecutor(connections[DEFAULT_DB_ALIAS])
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tim/code/django/django/db/migrations/executor.py", line 18, in __init__
    self.loader = MigrationLoader(self.connection)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tim/code/django/django/db/migrations/loader.py", line 58, in __init__
    self.build_graph()
  File "/home/tim/code/django/django/db/migrations/loader.py", line 235, in build_graph
    self.applied_migrations = recorder.applied_migrations()
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tim/code/django/django/db/migrations/recorder.py", line 89, in applied_migrations
    if self.has_table():
       ^^^^^^^^^^^^^^^^
  File "/home/tim/code/django/django/db/migrations/recorder.py", line 64, in has_table
    tables = self.connection.introspection.table_names(cursor)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tim/code/django-mongodb/django_mongodb/introspection.py", line 10, in table_names
    return sorted([x["name"] for x in self.connection.database.list_collections()])
                                      ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tim/code/django-mongodb/django_mongodb/base.py", line 150, in __getattr__
    return getattr(self, attr)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/tim/code/django-mongodb/django_mongodb/base.py", line 150, in __getattr__
    return getattr(self, attr)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/tim/code/django-mongodb/django_mongodb/base.py", line 150, in __getattr__
    return getattr(self, attr)
           ^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 741 more times]
RecursionError: maximum recursion depth exceeded
```

tests/utils_/test_parse_uri.py
aclark4life: Yes, but only on one of the tests, e526ab995b5b6a0b1ed61b7cca05d39ebc2906b1

tests/utils_/test_parse_uri.py
aclark4life: That's the bug I found a while back! OK I put it back 56fe609393d8cde55b466f1f6429ce4d12460566

tests/utils_/test_parse_uri.py
timgraham: Please add some explanatory comments about the reason mocking is needed.

README.md
```diff
@@ -121,6 +121,23 @@ DATABASES = {
...
+
+MONGODB_URI = "mongodb+srv://myDatabaseUser:D1fficultP%40ssw0rd@cluster0.example.mongodb.net/myDatabase?retryWrites=true&w=majority&tls=false"
+DATABASES["default"] = django_mongodb.parse_uri(MONGODB_URI)
+```
+
+#### `django_mongodb.parse_uri(uri, conn_max_age=0, conn_health_checks=False, test=None)`
+
+- Use `conn_max_age` and `conn_health_checks` to configure [persistent database
+  connections](https://docs.djangoproject.com/en/stable/ref/databases/#persistent-database-connections).
+- Use `test` to provide a dictionary of [settings for test databases](
```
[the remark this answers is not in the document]
aclark4life: Errr, 4d155ac084f13246374142b70c8d429230471173

tests/utils_/test_parse_uri.py
aclark4life: Done in 6ec2d516255d59ae408daa4d20ea978c8b73ec11

README.md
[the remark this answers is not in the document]
timgraham: Alex, your first intuition is correct. These variables control Django behavior, not the database driver.

Persistent connections are supported:
https://github.com/django/django/blob/ded485464214a3f69b64402b7d82221279f80008/django/db/backends/base/base.py#L249-L250
https://github.com/django/django/blob/ded485464214a3f69b64402b7d82221279f80008/django/db/backends/base/base.py#L615-L617

Health check don't work because we don't use the cursor API: 
https://github.com/django/django/blob/ded485464214a3f69b64402b7d82221279f80008/django/db/backends/base/base.py#L248
(`backends.base.test_base.ConnectionHealthChecksTests` is skipped)

tests/utils_/test_parse_uri.py
timgraham: Ok, now I can reproduce the crash by removing  `@patch`. But that seems sufficient... as far as I can tell, `setUp()` is useless. I think it would be slightly more clear if the comment/patch were inside the test method:
```
        with patch("dns.resolver.resolve"):
            settings_dict = parse_uri(URI)
```

README.md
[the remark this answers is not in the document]
aclark4life: 0501907c71ed4a1b1ff74a04bb380d20d3cb6784

tests/utils_/test_parse_uri.py
aclark4life: Well played, 1d7e4456af5a8271ccdc6c02b40d3c294619895d, 3549729392a2c0e88e6699e0516980a4b69e332c

django_mongodb/utils.py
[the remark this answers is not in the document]
aclark4life: @timgraham Any thoughts on this or rather anything for me to do here per @Jibola ?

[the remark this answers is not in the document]
timgraham: I would add a test:
```python
    def test_hosts_without_ports(self):
        settings_dict = parse_uri(
            "mongodb://host1.net,host2.net/myDatabase"
        )
        self.assertEqual(settings_dict["HOST"], "host1.net:27017,host2.net:27017")
        self.assertEqual(settings_dict["PORT"], None)
```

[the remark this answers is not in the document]
aclark4life: Thanks, 5d6c38e70037744de9782fda3571120ae6368914

tests/utils_/test_parse_uri.py
timgraham: Jib said, "I think it's best to still have the name as a passed through kwarg because Atlas URIs do often not provide a database name." By reading the documentation, I hope the user should be able to understand they can append the database name to their URL. Offering a separate keyword increases complexity and makes it possible for them to provide two conflicting values.

aclark4life: `NAME` has to be set in `DATABASES` else Django crashes. Once #196 is merged that will be enforced by `get_connect_params`.  So end users that try to do `parse_uri("mongodb://localhost")` (or something like it, without setting `NAME`) will get an error from `get_connect_params`.

Our `parse_uri` is inspired by `dj-database-url` which is inspired by [The Twelve-Factor App](https://12factor.net/backing-services) which prioritizes what @timgraham was arguing about with `HOST`. `HOST` should contain a `HOST` and nothing else so that:

Via  [Twelve-Factor App](https://12factor.net/backing-services)

> Resources can be attached to and detached from deploys at will. For example, if the app’s database is misbehaving due to a hardware issue, the app’s administrator might spin up a new database server restored from a recent backup. The current production database could be detached, and the new database attached – all without any code changes.

`dj-database-url` set the precedent that additional `kwargs` passed into `parse_uri` are passed to the `DATABASES` dictionary and `NAME` has already been parsed in the URL, so we definitely would not consider adding an additional `kwarg` for `NAME`. As @timgraham says, just direct users to the documentation that says "The database `NAME` is part of the URL" per  [Twelve-Factor App](https://12factor.net/backing-services).

timgraham: Reading https://www.mongodb.com/docs/manual/reference/connection-string/, one thing that could be confusing is that the example connection URL is: `mongodb+srv://[username:password@]host[/[defaultauthdb][?options]]`.

Does it imply that the authentication database could be different from the database that the user may want to store their Django models in? Reading https://pymongo.readthedocs.io/en/stable/examples/authentication.html#default-database-and-authsource, it looks like `defaultauthdb` can actually be the default database and you can specify a different authentication db using `authSource` in the query string. So I think our implementation is okay.

aclark4life: Right, I think it's OK too, but playing devil's advocate for @Jibola , Django says `NAME` [can be an empty string](https://docs.djangoproject.com/en/5.1/ref/settings/#name) and if we assume that MongoDB can auto-assign a collection name then we're back to the logistics of supporting an empty string `NAME` which involves an alternative fix for #196 and additional fixes for `parse_uri`. I don't have a strong preference either way and may even lean towards supporting the empty string `NAME` but I'll pass back to @Jibola for additional feedback.

I will note though that even though Django says `NAME` can be an empty string, that appears just a practical matter for Django to leave it to the back end to decide. If your database is SQLite and your `NAME` is empty what happens? I think SQLite picks a file name. If your database is Postgresql and your `NAME` is empty what happens? I think you get an error, but I'm not sure. Same with MySQL. If the MongoDB preference is for `NAME` to be an empty string and there is no error, then I imagine we can get Django to play along with that.

timgraham: Django's documentation says the the default of NAME is an empty string, but that doesn't mean it's a valid configuration. Some database backends like sqlite3 [raise an error if NAME isn't provided](https://github.com/django/django/blob/5e998d717f7b4220a1728bd49b66ca0162e2a6cb/django/db/backends/sqlite3/base.py#L157-L161), others like have the notion of [a "no db" connection](https://github.com/django/django/blob/5e998d717f7b4220a1728bd49b66ca0162e2a6cb/django/db/backends/postgresql/base.py#L240-L264) (`'NAME': None`) that is of some limited use. For example, when testing on PostgreSQL, "Normally Django will use a connection to the 'postgres' database to avoid running initialization queries against the production."

If there are cases where using Django + MongoDB without a database name is a valid configuration, we need the details. The behavior I see is:
```
>>> db_name = ''
>>> MongoClient(...)[db_name]
...
  File ".../pymongo/synchronous/mongo_client.py", line 1183, in __getitem__
    return database.Database(self, name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../pymongo/synchronous/database.py", line 133, in __init__
    _check_name(name)
  File ".../pymongo/database_shared.py", line 27, in _check_name
    raise InvalidName("database name cannot be the empty string")
pymongo.errors.InvalidName: database name cannot be the empty string
```

aclark4life: Well played

README.md
```diff
@@ -110,17 +110,57 @@ to this:
 DATABASES = {
     "default": {
         "ENGINE": "django_mongodb",
+        "HOST": "mongodb+srv://cluster0.example.mongodb.net",
```
timgraham, line 113: Is it correct to say that the scheme is required in HOST only for srv connections? (i.e. `mongodb://` is never required -- as tests seem to indicate?)

aclark4life: Yes, and in `parse_uri` I think we only add scheme to host if srv connection.

timgraham: I will add: 
```
`HOST` only needs a scheme prefix for SRV connections (`mongodb+srv://`). A
`mongodb://` prefix is never required.
```

timgraham (approved):

PR #238, opened 2025-01-29
PYTHON-5047 Fix dry run logic in releases again

blink1073:


Jibola:
The test is currently failing because of the 10.10.10.10 issue.

PR #240, opened 2025-01-29
PYTHON-5047 Fix dry run logic in releases yet again

blink1073:
See https://github.com/mongodb/motor/pull/323 for details.

### DIFF
--- a/.github/workflows/release-python.yml
+++ b/.github/workflows/release-python.yml
@@ -23,7 +23,9 @@ env:
   SILK_ASSET_GROUP: django-mongodb-backend-main
   EVERGREEN_PROJECT: django-mongodb-backend
   # Constant
-  DRY_RUN: ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run || 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-python.yml
```diff
@@ -23,7 +23,9 @@ env:
   SILK_ASSET_GROUP: django-mongodb-backend-main
   EVERGREEN_PROJECT: django-mongodb-backend
   # Constant
-  DRY_RUN: ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run || 'true' }}
+  # inputs will be empty on a scheduled run.  so, we only set dry_run
+  # to 'false' when the input is set to 'false'.
+  DRY_RUN: ${{ ! contains(inputs.dry_run, 'false') }}
```
Jibola, line 28: why does it have to be a contains?
Would an equality not work the same?

Jibola (approved):
Non-blocking question

.github/workflows/release-python.yml
blink1073: Perhaps, but this variant has already been tested. ;)

PR #264, opened 2025-03-03
Add SearchIndex and VectorSearchIndex

WaVEV:


django_mongodb_backend/schema.py
```diff
@@ -260,47 +261,33 @@ def alter_unique_together(
                 model, constraint, parent_model=parent_model, column_prefix=column_prefix
             )
 
+    @singledispatchmethod
```
timgraham, line 264: I've never seen this pattern before. My first reaction is that it less readable than:
```python
   def _add_index(self, index, model):
        if isinstance(index, SearchIndexModel):
             ...
        else isinstance(...)
            ....
        raise ValueError(f"{type(index)} isn't a supported index type")
```
Do you disagree?

django_mongodb_backend/base.py
```diff
@@ -127,6 +127,38 @@ def _isnull_operator(a, b):
...
+        "PositiveBigIntegerField": "number",
+        "PositiveIntegerField": "number",
+        "PositiveSmallIntegerField": "number",
+        "SlugField": "string",
+        "SmallAutoField": "number",
+        "SmallIntegerField": "number",
+        "TextField": "string",
+        "TimeField": "date",
+        "UUIDField": "uuid",
+        "ObjectIdAutoField": ["objectId", "number"],
```
timgraham, line 157: I assume you started this patch before 59ace002ab3b976d92690f4532e8ed2af53c5aea.

django_mongodb_backend/base.py
```diff
@@ -127,6 +127,38 @@ def _isnull_operator(a, b):
         "iregex": lambda a, b: regex_match(a, b, insensitive=True),
     }
 
+    mongo_data_types = {
```
timgraham, line 130: I guess this is needs to be different than `DatabaseWrapper.data_types` but we need a comment explaining the distinction.

django_mongodb_backend/base.py
```diff
@@ -175,7 +207,7 @@ def get_connection_params(self):
 
     @async_unsafe
     def get_new_connection(self, conn_params):
-        return MongoClient(**conn_params, driver=self._driver_info())
+        return MongoClient(**conn_params, directConnection=True, driver=self._driver_info())
```
timgraham, line 210: MongoDB team members can better say if we're okay to hardcode this option, but if so, I imagine it may need to be documented (i.e. is https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#directconnection-defaults-to-false of any concern?).

requirements.txt
```diff
@@ -1,2 +1,2 @@
 django>=5.1,<5.2
-pymongo>=4.6,<5.0
+pymongo>=4.11,<5.0
```
timgraham, line 2: Perhaps this is fine, but just want to note it's cutting off 2 years of versions (Nov 1, 2023 to Jan 28, 2025).

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+        with connection.schema_editor() as editor:
+            msg = "Article has no field named 'number1'"
+            with self.assertRaisesMessage(
+                FieldDoesNotExist, msg
+            ), connection.schema_editor() as editor:
+                editor.add_index(index=index, model=Article)
+
+
+class AtlasIndexTestsWithData(AtlasIndexTests):
```
timgraham, line 98: How does data affect the test? I'd think clearly something like `test_field_not_exists` doesn't need to be run in both cases.

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,218 @@
...
+            )
+            for i in range(5)
+        ]
+        cls.objs = Article.objects.bulk_create(articles)
+
+
+class AtlasSearchIndexTests(TestCase):
+    # Schema editor is used to  create the index to test that it works.
+    # available_apps = ["indexes"]
+    available_apps = None  # could be removed?
```
timgraham, line 122: It's not currently that helpful, but I'd include it for future possibilities. Without it, if we try to run Django's test suite without arguments, Django has to flush all test apps between each test (see  86d5885b34855a123756462f9d6954eb9420f763). If we added `available_apps` to all of Django's `TestCase` classes, then it would be feasible to run the Django test suite (without any test app names).

django_mongodb_backend/schema.py
WaVEV: Well, I discovered it few months ago, I have no preference, also the explicit way is easy to track, the other way is more indirect. Sometimes, the dispatcher is far from the main method making difficult to track.

Will change

django_mongodb_backend/base.py
WaVEV: 😬 Yes, now the number is allowed. At least in my mind.

django_mongodb_backend/base.py
WaVEV: It's hardcoded because the image that I am using requires it, it could be in the options or somewhere.

requirements.txt
WaVEV: With 4.6 didn't work. But in 4.7 it works fine. So I will change it to 4.7

django_mongodb_backend/introspection.py
```diff
@@ -30,3 +32,32 @@ def get_constraints(self, cursor, table_name):
                 "options": {},
             }
         return constraints
+
+    def _get_atlas_index_info(self, table_name):
```
timgraham, line 36: I wonder if we want the "atlas" name here when all of pymongo's APIs don't include it (`list_search_indexes()`, etc.). It also applies to the name of the Django index classes vs. pymongo's SearchIndexModel.

django_mongodb_backend/schema.py
```diff
@@ -260,47 +259,28 @@ def alter_unique_together(
                 model, constraint, parent_model=parent_model, column_prefix=column_prefix
             )
 
+    def _add_index(self, index, model):
```
timgraham, line 262: Do we need a separate method for _add/_drop_index?

django_mongodb_backend/base.py
timgraham: Would it be better organized in the index class or do you see some future use of it in other areas? I don't want to make things too tricky, but perhaps a mapping of `data_types` values to these search data types would simply this a bit. (e.g. string -> string, int -> number, with some exceptions....)

WaVEV: I think this is only used in atlas. The types that are described in base. So I agree to move this mapping in the index class.

django_mongodb_backend/introspection.py
WaVEV: Yes, it sounds logical. Will change atlas for search.

django_mongodb_backend/schema.py
WaVEV: Nope, isn't needed anymore

django_mongodb_backend/base.py
WaVEV: To the other question, it will work with something like double mapping with exceptions. 🤔 but it should be a function/method instead of a dictionaries.

tests/indexes_/test_atlas_indexes.py
WaVEV: After testing and figuring out that data is an unattended condition. I will remove those tests.

django_mongodb_backend/base.py
WaVEV: If we do a double mapping, connection should be passed to the function. So, it is not bad to have a double mapping but it will cost to add a new parameter. I would like to keep this as it is, maybe the dictionary is long but the flow is more simple. (given we don't need to pass connection)

django_mongodb_backend/indexes.py
```diff
@@ -58,7 +75,176 @@ def where_node_idx(self, compiler, connection):
     return mql
 
 
+def create_mongodb_index(
```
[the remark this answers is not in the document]
timgraham: How about calling this something like `get_pymongo_index_model()`? I think putting "create" in the name isn't intuitive because it doesn't create the index in the database. Also, could you put this refactor in a separate PR to precede the new feature so that those changes are mixed in with the new feature. Should make review and git blame easier.

django_mongodb_backend/indexes.py
```diff
@@ -58,7 +75,176 @@ def where_node_idx(self, compiler, connection):
...
+                BooleanField
+                | IntegerField
+                | DateField
+                | DateTimeField
+                | CharField
+                | TextField
+                | UUIDField
+                | ObjectIdField
+                | ObjectIdAutoField,
+            ):
```
timgraham, line 237: Is there any documentation you can link to that explains this list?

django_mongodb_backend/indexes.py
```diff
@@ -58,7 +75,176 @@ def where_node_idx(self, compiler, connection):
...
+    def __init__(self, *expressions, **kwargs):
+        super().__init__(*expressions, **kwargs)
+
+    def create_mongodb_index(
+        self, model, schema_editor, field=None, unique=False, column_prefix=""
+    ):
+        fields = {}
+        for field_name, _ in self.fields_orders:
+            field_ = model._meta.get_field(field_name)
+            type_ = self.search_index_data_types[field_.get_internal_type()]
```
timgraham, line 170: You can use `schema_editor.connection` here if that's where you need it for the mapping.

django_mongodb_backend/indexes.py
```diff
@@ -58,7 +75,176 @@ def where_node_idx(self, compiler, connection):
...
+            field_path = column_prefix + model._meta.get_field(field_name).column
+            fields[field_path] = {"type": type_}
+        return SearchIndexModel(
+            definition={"mappings": {"dynamic": False, "fields": fields}}, name=self.name
+        )
+
+
+class VectorSearchIndex(Index):
+    suffix = "vector_search"
+    ALLOWED_SIMILARITY_FUNCTIONS = ("euclidean", "cosine", "dotProduct")
```
timgraham, line 180: This could be a set literal.

django_mongodb_backend/indexes.py
```diff
@@ -58,7 +75,176 @@ def where_node_idx(self, compiler, connection):
...
+            )
+            for field_name, order in self.fields_orders
+        ]
+    )
+    return IndexModel(index_orders, name=self.name, **kwargs)
+
+
+class SearchIndex(Index):
+    suffix = "search"
+    # Maps Django internal type to atlas search index type.
```
timgraham, line 128: Is there a documentation reference for this list?

django_mongodb_backend/fields/array.py
```diff
@@ -27,14 +28,23 @@ class ArrayField(CheckFieldDefaultMixin, Field):
     }
     _default_hint = ("list", "[]")
 
-    def __init__(self, base_field, size=None, **kwargs):
+    def __init__(self, base_field, size=None, fixed_size=None, **kwargs):
```
timgraham, line 31: The parameter names don't look clear. I'm inclined to rename `size` to `max_size` and `fixed_size` to `size`. What do you think? Let's add this feature in a separate PR to precede this feature. I'm thinking first commit: rename `size` to `max_size`, second commit: add `size`.

django_mongodb_backend/schema.py
```diff
@@ -264,43 +263,21 @@ def alter_unique_together(
...
-        idx = IndexModel(index_orders, name=index.name, **kwargs)
-        model = parent_model or model
-        self.get_collection(model._meta.db_table).create_indexes([idx])
+        if idx:
+            model = parent_model or model
+            if isinstance(idx, SearchIndexModel):
+                self.get_collection(model._meta.db_table).create_search_index(idx)
+            elif isinstance(idx, IndexModel):
+                self.get_collection(model._meta.db_table).create_indexes([idx])
+            else:
```
timgraham, line 279: I think we shouldn't add exceptions that our code can't get to. I would write:
```
else:
    self.get_collection(model._meta.db_table).create_indexes([idx])
```

django_mongodb_backend/indexes.py
WaVEV: You are right, 🤔 ok, I will try the double mapping and see how it goes.

django_mongodb_backend/indexes.py
WaVEV: Yes, it's [here](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#atlas-vector-search-index-fields). It is in fields.type table entry. I think copying the exact text: 
```text
filter - for fields that contain boolean, date, objectId, numeric, string, or UUID values.
```
Maybe we could make use the type mapping here? Instead of enumerating all the supporting fields?

timgraham: It seems type checking may be more extensible, e.g. for custom fields.

django_mongodb_backend/schema.py
```diff
@@ -318,7 +293,12 @@ def _add_field_index(self, model, field, *, column_prefix=""):
     def remove_index(self, model, index):
         if index.contains_expressions:
             return
-        self.get_collection(model._meta.db_table).drop_index(index.name)
+        if isinstance(index, SearchIndex | VectorSearchIndex):
+            self.get_collection(model._meta.db_table).drop_search_index(index.name)
+        elif isinstance(index, Index):
+            self.get_collection(model._meta.db_table).drop_index(index.name)
+        else:
+            raise ValueError(f"{type(index)} isn't a supported index type")
```
timgraham, line 301: as above

django_mongodb_backend/indexes.py
timgraham: The link would be okay.

django_mongodb_backend/indexes.py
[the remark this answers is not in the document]
WaVEV: work on #272

django_mongodb_backend/indexes.py
WaVEV: 🤔 ok, so, checking by the internal type is better and more flexible .

django_mongodb_backend/schema.py
```diff
@@ -264,43 +263,21 @@ def alter_unique_together(
...
+            unique=unique,
+            column_prefix=column_prefix,
         )
-        idx = IndexModel(index_orders, name=index.name, **kwargs)
-        model = parent_model or model
-        self.get_collection(model._meta.db_table).create_indexes([idx])
+        if idx:
+            model = parent_model or model
+            if isinstance(idx, SearchIndexModel):
+                self.get_collection(model._meta.db_table).create_search_index(idx)
```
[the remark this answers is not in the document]
WaVEV: Added

django_mongodb_backend/indexes.py
```diff
@@ -58,7 +69,153 @@ def where_node_idx(self, compiler, connection):
...
+            field_ = model._meta.get_field(field_name)
+            field_path = column_prefix + model._meta.get_field(field_name).column
+            mappings = {"path": field_path}
+            if isinstance(field_, ArrayField):
+                try:
+                    vector_size = int(field_.size)
+                except (ValueError, TypeError) as err:
+                    raise ValueError("Atlas vector search requires size.") from err
+                if not isinstance(field_.base_field, FloatField | DecimalField):
+                    raise ValueError("Base type must be Float or Decimal.")
```
timgraham, line 194: This sort of validation really belongs as a system check, since at this point the user has already generated an invalid migration and has to do quite a lot to fix the problem. Unfortunately, the `Index.check()` hook [is missing](https://code.djangoproject.com/ticket/36273#ticket). Until then, however, I think we can [write a system check](https://docs.djangoproject.com/en/dev/topics/checks/#writing-your-own-checks) that iterates over all the models in `app_configs` and does this validation. Let me know if you run into trouble with that idea.

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,204 @@
...
+            editor.add_index(index=index, model=Article)
+
+    def test_field_size_required(self):
+        index = VectorSearchIndex(
+            name="recent_article_idx",
+            fields=["number_list"],
+        )
+        msg = "Atlas vector search requires size."
+        with self.assertRaisesMessage(ValueError, msg), connection.schema_editor() as editor:
+            editor.add_index(index=index, model=Article)
```
timgraham, line 176: For testing invalid model logic (which can hopefully be done as system checks), see
https://github.com/django/django/blob/922c1c732a47c02aa5ef28b0b1a2bd9bc9b92d87/tests/check_framework/test_model_checks.py

django_mongodb_backend/indexes.py
```diff
@@ -58,7 +69,153 @@ def where_node_idx(self, compiler, connection):
...
+    suffix = "vector_search"
+    ALLOWED_SIMILARITY_FUNCTIONS = frozenset(("euclidean", "cosine", "dotProduct"))
+
+    def __init__(self, *expressions, similarities="cosine", **kwargs):
+        super().__init__(*expressions, **kwargs)
+        # validate the similarities types
+        if isinstance(similarities, str):
+            self._check_similarity_functions([similarities])
+        else:
+            self._check_similarity_functions(similarities)
```
timgraham, line 164: I'm not sure I like accepting both a string and a list. Did you find a precedent for that? Even if so, `self.similarities` could at least be stored as a list to simplify future usage. Since you have a custom parameter, you need to add a `deconstruct()` method.

django_mongodb_backend/fields/array.py
timgraham: I did part 1 in https://github.com/mongodb/django-mongodb-backend/pull/273. Want to send a PR to my branch with part 2? There is a bit more to do than you already did (like handle the parameter in `deconstruct()`, add it to test migrations, support in forms, and update documentation). Hopefully my PR gives hints about the places that need updating.

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,204 @@
...
+            self.assertAddRemoveIndex(editor, Article, index)
+
+    def test_field_not_exists(self):
+        index = SearchIndex(
+            name="recent_article_idx",
+            fields=["headline", "non_existing_name"],
+        )
+        msg = "Article has no field named 'non_existing_name'"
+        with self.assertRaisesMessage(FieldDoesNotExist, msg), connection.schema_editor() as editor:
+            editor.add_index(index=index, model=Article)
```
timgraham, line 89: This test may be unnecessary since it's already validated by a system check:
```
search_indexes.Article: (models.E012) 'indexes' refers to the nonexistent field 'headline'.
```

django_mongodb_backend/indexes.py
WaVEV: I tried the pattern: 
single value applies to all elements, and a list assigns values one-to-one.
Like:
[border parameter of expand in pillow](https://pillow.readthedocs.io/en/stable/reference/ImageOps.html#PIL.ImageOps.expand)
But I can make it one to one.

timgraham: I didn't realize that, so I'll look again.

django_mongodb_backend/fields/array.py
WaVEV: Ok, shall I add those change also to my [PR](https://github.com/timgraham/django-mongodb/pull/8)?

django_mongodb_backend/indexes.py
```diff
@@ -1,12 +1,20 @@
...
+import itertools
+from collections import defaultdict
 
 from django.db import NotSupportedError
-from django.db.models import Index
+from django.db.models import (
+    DecimalField,
+    FloatField,
+    Index,
+)
```
timgraham, line 10: Items were removed, so this can be a single line.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +109,156 @@ def where_node_idx(self, compiler, connection):
     return mql
 
 
+def get_pymongo_index_model(
```
timgraham, line 112: This is added redundantly after the rebase.

django_mongodb_backend/indexes.py
WaVEV: I am creating the system check, the problem here is: can I import connection. Can I get that?

.github/workflows/mongodb_settings.py
```diff
@@ -17,10 +17,12 @@
         "default": {
             "ENGINE": "django_mongodb_backend",
             "NAME": "djangotests",
+            "OPTIONS": {"directConnection": True},
```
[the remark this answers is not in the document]
WaVEV: [Docker site](https://hub.docker.com/r/mongodb/mongodb-atlas-local) don't provide much info about it. Just give the connection uri format.
I think is to connect to a single MongoDB instance like standalone mode. But need to confirm that.

django_mongodb_backend/fields/array.py
```diff
@@ -27,13 +28,21 @@ class ArrayField(CheckFieldDefaultMixin, Field):
     }
     _default_hint = ("list", "[]")
 
-    def __init__(self, base_field, size=None, **kwargs):
+    def __init__(self, base_field, max_size=None, size=None, **kwargs):
```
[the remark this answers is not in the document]
WaVEV: Array is not only for vector search, is for general purpose. Like array of array. Or array of char field or array of embedded fields

django_mongodb_backend/indexes.py
WaVEV: A first approach to this is implemented.

django_mongodb_backend/fields/validators.py
```diff
@@ -0,0 +1,19 @@
+from django.core.validators import BaseValidator
+from django.utils.deconstruct import deconstructible
+from django.utils.translation import ngettext_lazy
+
+
+@deconstructible
+class LengthValidator(BaseValidator):
```
timgraham, line 7: Chop. (Moved to top-level `validators.py`.)

django_mongodb_backend/checks.py
```diff
@@ -0,0 +1,27 @@
...
+        models = apps.get_models()
+    else:
+        models = chain.from_iterable(app_config.get_models() for app_config in app_configs)
+    for model in models:
+        for db in databases or ():
+            if not router.allow_migrate_model(db, model):
+                continue
+            connection = connections[db]
+            for model_index in model._meta.indexes:
+                if not isinstance(model_index, VectorSearchIndex):
```
timgraham, line 24: I'd favor `hasattr(model_index, "check")`

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,136 @@ def where_node_idx(self, compiler, connection):
...
+        )
+
+
+class VectorSearchIndex(SearchIndex):
+    suffix = "vsi"
+    ALLOWED_SIMILARITY_FUNCTIONS = frozenset(("euclidean", "cosine", "dotProduct"))
+
+    def __init__(self, *expressions, similarities="cosine", **kwargs):
+        super().__init__(*expressions, **kwargs)
+        # validate the similarities types
```
timgraham, line 150: chop comment

tests/indexes_/models.py
```diff
@@ -1,7 +1,21 @@
...
     headline = models.CharField(max_length=100)
     number = models.IntegerField()
     body = models.TextField()
+    data = models.JSONField()
+    embedded = EmbeddedModelField(Data)
+    auto_now = models.DateTimeField(auto_now=True)
+    title_embedded = ArrayField(models.FloatField(), size=10)
+    description_embedded = ArrayField(models.FloatField(), size=10)
+    number_list = ArrayField(models.FloatField())
+    name_list = ArrayField(models.CharField(max_length=30), size=10)
```
timgraham, line 21: Some fields that are moved to invalid model checks can be removed.

tests/indexes_/models.py
```diff
@@ -1,7 +1,21 @@
...
+    integer = models.IntegerField()
+
 
 class Article(models.Model):
     headline = models.CharField(max_length=100)
     number = models.IntegerField()
     body = models.TextField()
+    data = models.JSONField()
+    embedded = EmbeddedModelField(Data)
+    auto_now = models.DateTimeField(auto_now=True)
```
timgraham, line 17: I'd suggest a more semantic field name like `created_at`.

tests/validators_/tests.py
```diff
@@ -0,0 +1,14 @@
+from django.core.exceptions import ValidationError
```
timgraham, line 1: Chop. (Moved to `test_length.py`.)

timgraham (commented):
Let's add a check similar to https://github.com/django/django/blob/1429e722f265a4f4229b5f7eaa6a6df3161c342a/django/db/models/constraints.py#L150-L167 and make sure that schema editor ignores search indexes if not supported.

django_mongodb_backend/features.py
```diff
@@ -611,3 +612,21 @@ def django_test_expected_failures(self):
     @cached_property
     def is_mongodb_6_3(self):
         return self.connection.get_database_version() >= (6, 3)
+
+    @cached_property
+    def supports_search_indexes(self):
```
ShaneHarvey, line 617: Do we need this property? Is it only added for the test suite?

timgraham: Besides skipping the relevant tests, I [also suggested](https://github.com/mongodb/django-mongodb-backend/pull/264#pullrequestreview-2753565747) using it to alert the user if they try to use an index that isn't supported.

WaVEV: I researched it and I didn't find a more viable way to check if the Atlas search is supported. 😬

timgraham: Will the same check work for detecting support for Atlas search queries? If so, maybe we want to call it something more generic like `supports_atlas_search` so we can also use the feature flag there.

WaVEV: Yes, 🤔 I wanted to avoid using atlas as a name but the feature is called [atlas search](https://www.mongodb.com/docs/atlas/atlas-search/). So yes, will rename.

ShaneHarvey: Based on my experience, this pattern of creating/dropping a collection is problematic. For test only code it's fine, but for apps we should use a different approach. One way would be to just surface the search/vector errors directly (or wrap them if they aren't informative enough).

WaVEV: We could create the collection if it isn't there and not remove it.
For the alternative way I am not getting the point. Currently, we are surfacing over an error when try to use an atlas operation, The  error is handled, and then we know if atlas is supported.

timgraham: If we need to support MongoDB 6 (pending discussion), we can create the collection only for that version and use `django_migrations` since this built-in collection doesn't require any special creation setups such as indexes.

.github/workflows/test-python.yml
```diff
@@ -50,8 +50,9 @@ jobs:
       - name: Copy the test runner file
         run: cp .github/workflows/runtests.py django_repo/tests/runtests_.py
       - name: Start MongoDB
-        uses: supercharge/mongodb-github-action@1.12.0
+        uses: wavev/mongodb-github-action@atlas
```
timgraham, line 53: Do we have to duplicate most of this entire script in order to also test against the existing non-Atlas MongoDB?

tests/system_checks/tests.py
```diff
@@ -1,19 +1,50 @@
...
 @override_system_checks([check_vector_search_indexes])
-class InvalidSearchIndexesTest(SimpleTestCase):
+class InvalidSearchIndexesTest(TestCase):
+    def test_search_requires_search_index_support(self):
+        class Article(models.Model):
+            title = models.CharField(max_length=10)
+
+            class Meta:
+                indexes = [
+                    SearchIndex(fields=["title"]),
```
timgraham, line 24: Could you please also add a VectorSearchIndex to verify that `super().check()` is called and that it works properly.

django_mongodb_backend/indexes.py
```diff
@@ -107,6 +107,23 @@ def where_node_idx(self, compiler, connection):
...
+
+    def check(self, model, connection):
+        errors = []
+        if not connection.features.supports_search_indexes:
+            errors.append(
+                Warning(
+                    "This version of MongoDB does not support search indexes.",
+                    hint=(
+                        "The index won't be created. Silence this warning if you "
+                        "don't care about it."
```
timgraham, line 120: I believe you need to add a check in SchemaEditor to ignore the index if unsupported.

django_mongodb_backend/features.py
WaVEV: is django_migrations a collection?

tests/system_checks/tests.py
```diff
@@ -0,0 +1,174 @@
...
+                    id="django_mongodb_backend.indexes.SearchIndex.W001",
+                )
+            ],
+        )
+
+
+@skipUnlessDBFeature("supports_atlas_search")
+@isolate_apps("system_checks", attr_name="apps")
+@override_system_checks([check_vector_search_indexes])
+class InvalidVectorSearchIndexesTest(TestCase):
```
timgraham, line 47: always "Tests" (plural)... some classes in Django are inconsistent.

tests/system_checks/tests.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_vectorsearch_requires_size(self):
```
timgraham, line 48: Since this is `InvalidVectorSearchIndexes`, we don't need the `vectorsearch` prefix in test names.

tests/indexes_/models.py
```diff
@@ -1,7 +1,19 @@
...
+
 
 class Article(models.Model):
     headline = models.CharField(max_length=100)
     number = models.IntegerField()
     body = models.TextField()
+    data = models.JSONField()
+    embedded = EmbeddedModelField(Data)
+    created_at = models.DateTimeField(auto_now=True)
+    title_embedded = ArrayField(models.FloatField(), size=10)
```
timgraham, line 18: title_embedded is unused

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,152 @@ def where_node_idx(self, compiler, connection):
...
+                )
+        for field_name, _ in self.fields_orders:
+            field_ = model._meta.get_field(field_name)
+            if isinstance(field_, ArrayField):
+                try:
+                    int(field_.size)
+                except (ValueError, TypeError):
+                    errors.append(
+                        Error(
+                            "Atlas vector search requires size.",
```
timgraham, line 193: "requires size on field <field_name>"

django_mongodb_backend/checks.py
```diff
@@ -0,0 +1,24 @@
+from itertools import chain
+
+from django.apps import apps
+from django.core.checks import Tags, register
+from django.db import connections, router
+
+
+@register(Tags.models)
```
timgraham, line 8: I think you need to register the check in `__init__.py` (perhaps following the existing pattern of creating a `register_checks()` function). The tests are working because of `override_system_checks()` but this may not be necessary after you fix this.

tests/system_checks/tests.py
```diff
@@ -0,0 +1,174 @@
+from django.core import checks
```
timgraham, line 1: I'd organize this in `indexes_/test_checks.py`

django_mongodb_backend/features.py
timgraham: https://github.com/django/django/blob/5020a9d43a9ebd5f20f9a77e5fed424023facb15/django/db/migrations/recorder.py#L32-L40

.github/workflows/test-python.yml
WaVEV: I don't know if I understood the question, but the action could handle both kind of MongoDB images, atlas and the regular one, if we migrate to a shell script, it also can handle it without much effort.
So we only need to add 2 steps more. One for start a MongoDB without atlas and run the test again. Maybe we could avoid running all the test, for atlas just run system_check and indexes_ and for the other regular server, run all the test suite.

[Like langchain does. ](https://github.com/langchain-ai/langchain-mongodb/blob/main/.github/workflows/_test.yml#L41-L63)

django_mongodb_backend/checks.py
WaVEV: Yes, I need. otherwise the code is unreachable, I cannot get rid of override_system_checks in test, because there are some other checks that are triggered. To isolate the test, I will keep the override_system_checks.

django_mongodb_backend/indexes.py
WaVEV: Maybe it should be in `get_pymongo_index_model`. Return None if the index cannot be made.

timgraham: Sure

django_mongodb_backend/schema.py
```diff
@@ -283,7 +292,17 @@ def _add_field_index(self, model, field, *, column_prefix=""):
     def remove_index(self, model, index):
         if index.contains_expressions:
             return
-        self.get_collection(model._meta.db_table).drop_index(index.name)
+        if isinstance(index, SearchIndex | VectorSearchIndex):
+            # Drop the index if it exists, particularly if it may not have been created previously
+            # due to lack of Atlas search support, but now the database supports it.
```
timgraham, line 297: Is it really possible/likely that a database can go from non-Atlas to Atlas?

tests/indexes_/test_checks.py
```diff
@@ -0,0 +1,205 @@
...
+                    id="django_mongodb_backend.indexes.SearchIndex.W001",
+                )
+            ],
+        )
+
+
+@isolate_apps("indexes_", attr_name="apps")
+@override_system_checks([check_vector_search_indexes])
+class InvalidVectorSearchIndexesTests(TestCase):
+    @skipIfDBFeature("supports_atlas_search")
```
timgraham, line 47: I don't like `skipUnlessDBFeature()` on every test. It would be simpler at the class level. I would move this test to `UnsupportedSearchIndexesTests` (renaming `InvalidSearchIndexesTests`) so you can do that.

tests/indexes_/test_checks.py
```diff
@@ -0,0 +1,205 @@
...
+    def test_requires_atlas_search_support(self):
+        class Article(models.Model):
+            title = models.CharField(max_length=10)
+
+            class Meta:
+                indexes = [
+                    VectorSearchIndex(fields=["title"]),
+                ]
```
timgraham, line 55: I think it's fine to put this sort of definition on one line if it's just a single index that's not too long.

tests/indexes_/models.py
```diff
@@ -1,7 +1,18 @@
...
+
 
 class Article(models.Model):
     headline = models.CharField(max_length=100)
     number = models.IntegerField()
     body = models.TextField()
+    data = models.JSONField()
+    embedded = EmbeddedModelField(Data)
+    created_at = models.DateTimeField(auto_now=True)
+    description_embedded = ArrayField(models.DecimalField(decimal_places=3, max_digits=10), size=10)
```
timgraham, line 18: Is the name `description_embedded` following a naming convention? Naively I might think "embedded" implies `EmbeddedModelField` so I'm just wondering if there's a naming convention I should be aware of.

WaVEV: ok, it is for the docs embedding. but we could just name it as vector, or semantic_vector

django_mongodb_backend/schema.py
WaVEV: maybe @Jibola could answer better, IMO it is very reasonable. An application add some support for AI and don't want to have two separate backends. The opposite isn't very reasonable, go from Atlas to non-Atlas. I decided to handle both.

django_mongodb_backend/schema.py
```diff
@@ -260,12 +263,18 @@ def alter_unique_together(
     def add_index(
         self, model, index, *, field=None, unique=False, column_prefix="", parent_model=None
     ):
+        # Check if the index can be created.
+        if hasattr(index, "check") and index.check(model, self.connection):
+            return
```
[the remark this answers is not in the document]
WaVEV: 🤔 well for the warning, check those flags is enough.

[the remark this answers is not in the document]
WaVEV: 🤔 The index should do the check and return None if cannot be made.

django_mongodb_backend/schema.py
timgraham: How do you migrate from non-Atlas to Atlas? Dump the db and load it in a new server? I guess it's possible for this migrations scenario to happen, but it doesn't seem very likely. I'd expect the main place where search indexes might be ignored is if they were in a third-party app, so basically the scenario would be: user installs third-party app with ignored search index, user migrates to atlas, third-party app removes atlas index. (Do you have another scenario in mind?)

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,169 @@ def where_node_idx(self, compiler, connection):
...
+        if not schema_editor.connection.features.supports_atlas_search:
+            return None
+        fields = {}
+        for field_name, _ in self.fields_orders:
+            field_ = model._meta.get_field(field_name)
+            type_ = self.search_index_data_types(field_, field_.db_type(schema_editor.connection))
+            field_path = column_prefix + model._meta.get_field(field_name).column
+            fields[field_path] = {"type": type_}
+        return SearchIndexModel(
+            definition={"mappings": {"dynamic": False, "fields": fields}}, name=self.name
```
timgraham, line 159: Was there discussion about supporting only [the expanded syntax](https://www.mongodb.com/docs/atlas/atlas-search/index-definitions/#syntax]) later? I'm thinking about what to write in the docs.

django_mongodb_backend/checks.py
```diff
@@ -0,0 +1,27 @@
+from itertools import chain
+
+from django.apps import apps
+from django.core.checks import Tags, register
+from django.db import connections, router
+
+
+def check_vector_search_indexes(app_configs, databases=None, **kwargs):  # noqa: ARG001
```
timgraham, line 8: We may add checks to other indexes, so I'd name it `check_indexes`.

.github/workflows/mongodb_settings.py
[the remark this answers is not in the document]
timgraham: I think Jib's suggestion wasn't to document what the parameter does (which might be fine), but that it's needed to connect to the Atlas image in docker.

In any case, I don't think we need to repeat the comment for the 'other' database.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,169 @@ def where_node_idx(self, compiler, connection):
...
+                            id=f"{self._error_id_prefix}.E002",
+                        )
+                    )
+            else:
+                field_type = field_.db_type(connection)
+                search_type = self.search_index_data_types(field_, field_type)
+                # filter - for fields that contain boolean, date, objectId,
+                # numeric, string, or UUID values. Reference:
+                # https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#atlas-vector-search-index-fields
+                if search_type not in ("number", "string", "boolean", "objectId", "uuid", "date"):
```
timgraham, line 227: use set literal rather than tuple

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,169 @@ def where_node_idx(self, compiler, connection):
...
+                        id=f"{self._error_id_prefix}.E004",
+                    )
+                )
+        viewed = set()
+        for field_name, _ in self.fields_orders:
+            if field_name in viewed:
+                errors.append(
+                    Error(
+                        f"Field '{field_name}' is defined more than once. Vector and filter"
+                        " fields must use distinct field names.",
```
timgraham, line 193: For consistency, in a multi-line string always put the space on the previous line.

django_mongodb_backend/schema.py
WaVEV: 🤔 yes, I think you are right. But at least the if must remains as it is. Or maybe we could remove the existence check. Because if an index was skipped because the backend doesn't support atlas, then It has nothing to drop.

tests/indexes_/test_checks.py
```diff
@@ -0,0 +1,223 @@
...
+                    id="django_mongodb_backend.indexes.SearchIndex.W001",
+                )
+            ],
+        )
+
+
+@skipIfDBFeature("supports_atlas_search")
+@isolate_apps("indexes_", attr_name="apps")
+@override_system_checks([check_vector_search_indexes])
+class UnsupportedSearchIndexesTests(TestCase):
```
timgraham, line 45: I meant to use this class name in the class above and combine the two.

tests/indexes_/test_checks.py
```diff
@@ -0,0 +1,223 @@
...
+
+            class Meta:
+                indexes = [VectorSearchIndex(fields=["title_embedded"])]
+
+        errors = checks.run_checks(app_configs=self.apps.get_app_configs(), databases={"default"})
+        self.assertEqual(
+            errors,
+            [
+                checks.Error(
+                    "Atlas vector search requires size on title_embedded.",
```
timgraham, line 86: You missed a word in my suggestion: "on field"

How about: "An Atlas vector search index requires you to specify Field.size on field Model.field_name."

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,229 @@
...
+        editor.remove_index(index=index, model=model)
+        self.assertNotIn(
+            index.name,
+            connection.introspection.get_constraints(
+                cursor=None,
+                table_name=model._meta.db_table,
+            ),
+        )
+
+    def test_drop_non_existing_index(self):
```
timgraham, line 136: The word you're looking for is "nonexistent".

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,229 @@
...
+            ),
+        )
+        editor.remove_index(index=index, model=model)
+        self.assertNotIn(
+            index.name,
+            connection.introspection.get_constraints(
+                cursor=None,
+                table_name=model._meta.db_table,
+            ),
+        )
```
timgraham, line 31: Use a TestMixin (as in (schema_/tests.py`) to share common assertions.

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,229 @@
...
+from django.db import connection
+from django.test import TestCase, skipIfDBFeature, skipUnlessDBFeature
+
+from django_mongodb_backend.indexes import SearchIndex, VectorSearchIndex
+
+from .models import Article
+
+
+class SearchIndexTests(TestCase):
+    # Tests for creating, validating, and removing search indexes using Django's schema editor.
```
timgraham, line 12: This looks like a docstring (as opposed to the comments in `schema_/test_embedded_model.py` which break up the tests into sections.

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,229 @@
...
+    available_apps = None
```
timgraham, line 13: unneeded

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,169 @@ def where_node_idx(self, compiler, connection):
...
+                        Error(
+                            f"Atlas vector search requires size on {field_name}.",
+                            obj=self,
+                            id=f"{self._error_id_prefix}.E001",
+                        )
+                    )
+                if not isinstance(field_.base_field, FloatField | DecimalField):
+                    errors.append(
+                        Error(
+                            "Base type must be Float or Decimal.",
```
timgraham, line 216: "An Atlas vector search index requires the base field of ArrayField Model.field_name to be FloatField or DecimalField but is <field_class>."

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +109,156 @@ def where_node_idx(self, compiler, connection):
...
+                    vector_size = int(field_.size)
+                except (ValueError, TypeError) as err:
+                    raise ValueError("Atlas vector search requires size.") from err
+                if not isinstance(field_.base_field, FloatField | DecimalField):
+                    raise ValueError("Base type must be Float or Decimal.")
+                mappings.update(
+                    {
+                        "type": "vector",
+                        "numDimensions": vector_size,
+                        "similarity": next(similarities),
```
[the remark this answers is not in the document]
timgraham: Do we need a check that the size of similarities (if a list) equals the number of fields?

django_mongodb_backend/indexes.py
WaVEV: Yes, I also remember that. When we asked about giving the user the full mapping or restricting it and creating the mapping under the hood.

django_mongodb_backend/indexes.py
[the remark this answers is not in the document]
WaVEV: 🤔 I let this without care. But maybe we should. will add

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,191 @@ def where_node_idx(self, compiler, connection):
...
+        for func in similarities:
+            if func not in self.ALLOWED_SIMILARITY_FUNCTIONS:
+                errors.append(
+                    Error(
+                        f"{func} isn't a valid similarity function, options "
+                        f"are {', '.join(sorted(self.ALLOWED_SIMILARITY_FUNCTIONS))}",
+                        obj=self,
+                        id=f"{self._error_id_prefix}.E004",
+                    )
+                )
```
timgraham, line 186: Anything that can be validated in `Index.__init__()` like similarities should be done there. That raises an error earlier in the process, especially because there's an issue in Django where `makemigrations` doesn't currently execute database checks like these. Validation that requires the model or a database connection can be left as a check.

.github/workflows/start_local_atlas.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -eu
+
+echo "Starting the container"
+
+IMAGE=mongodb/mongodb-atlas-local:latest
```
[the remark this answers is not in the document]
WaVEV: Ok, I can make this script with a parameter. And for now using 7.0.4 or 7.0.11

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,194 @@
...
+                cursor=None,
+                table_name=model._meta.db_table,
+            ),
+        )
+
+
+class SearchIndexTests(TestMixin, TestCase):
+    """Tests for creating, validating, and removing search indexes using Django's schema editor."""
+
+    @skipUnlessDBFeature("supports_atlas_search")
```
timgraham, line 32: As commented on other file, move skip decorator to the class.

tests/indexes_/test_atlas_indexes.py
```diff
@@ -0,0 +1,194 @@
...
+            self.assertAddRemoveIndex(editor, Article, index)
+
+
+class VectorSearchIndexTests(TestMixin, TestCase):
+    """
+    Tests for creating, validating, and removing vector search indexes
+    using Django's schema editor.
+    """
+
+    def test_invalid_similarity_function(self):
```
timgraham, line 107: Some tests like this one and deconstruct can be moved to a `SimpleTestCase` which is a little faster and doesn't need the db skip.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,183 @@ def where_node_idx(self, compiler, connection):
...
+                    Error(
+                        f"Field '{field_name}' is defined more than once. Vector and filter "
+                        "fields must use distinct field names.",
+                        obj=self,
+                        hint="If you need different configurations for the same field, "
+                        "create separate indexes.",
+                        id=f"{self._error_id_prefix}.E004",
+                    )
+                )
+                continue
```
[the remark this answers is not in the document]
WaVEV: Perfect,  I think `Atlas vector search requires size` must be in the __init__ too.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,183 @@ def where_node_idx(self, compiler, connection):
...
+            definition={"mappings": {"dynamic": False, "fields": fields}}, name=self.name
+        )
+
+
+class VectorSearchIndex(SearchIndex):
+    suffix = "vsi"
+    ALLOWED_SIMILARITY_FUNCTIONS = frozenset(("euclidean", "cosine", "dotProduct"))
+    _error_id_prefix = "django_mongodb_backend.indexes.VectorSearchIndex"
+
+    def __init__(self, *expressions, similarities="cosine", **kwargs):
```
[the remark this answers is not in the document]
WaVEV: the other index haven't got any __init__. It uses the same as normal index uses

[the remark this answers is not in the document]
timgraham: My thought was to add `SearchIndex.__init__()` with the more limited arguments to help users catch usage mistakes.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,185 @@ def where_node_idx(self, compiler, connection):
...
+        errors = []
+        if not connection.features.supports_atlas_search:
+            errors.append(
+                Warning(
+                    "This MongoDB server does not support atlas search.",
+                    hint=(
+                        "The index won't be created. Silence this warning if you "
+                        "don't care about it."
+                    ),
+                    obj=self,
```
timgraham, line 122: While working on [the hook in Django](https://github.com/django/django/pull/19395), I noticed the built-in checks provide `obj=model` which may be more helpful for locating the problematic index. But perhaps the warning message would also be more clear if it said: ``f"This MongoDB server does not support {self.__class__.__name__}."`

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,185 @@ def where_node_idx(self, compiler, connection):
...
+    _error_id_prefix = "django_mongodb_backend.indexes.SearchIndex"
+
+    def check(self, model, connection):
+        errors = []
+        if not connection.features.supports_atlas_search:
+            errors.append(
+                Warning(
+                    "This MongoDB server does not support atlas search.",
+                    hint=(
+                        "The index won't be created. Silence this warning if you "
```
timgraham, line 119: How about amending the second sentence: "Use an Atlas-enabled version of MongoDB, or silence..."

django_mongodb_backend/indexes.py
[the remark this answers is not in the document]
timgraham: I don't think so since that's a relational check that needs `model`.

django_mongodb_backend/indexes.py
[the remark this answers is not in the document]
WaVEV: So the kwargs must be removed?

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,180 @@ def where_node_idx(self, compiler, connection):
...
+            )
+        return errors
+
+    def search_index_data_types(self, field, db_type):
+        """
+        Map a model field's internal type to search index type.
+        Reference: https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#data-types
+        """
+        if field.get_internal_type() == "UUIDField":
+            return "uuid"
```
timgraham, line 134: As currently implemented, I believe this should be string: https://github.com/mongodb/django-mongodb-backend/blob/43f477cc2e1a0981e53bc2a60ee0e049939c716f/django_mongodb_backend/base.py#L62

Separately, maybe it's worth confirming that `UUIDField` can't store it values as BSON uuid. Setting [`DatabaseFeatures.has_native_uuid_field = True`](https://github.com/django/django/blob/main/django/db/models/fields/__init__.py#L2753-L2761) raises the error: ValueError: cannot encode native uuid.UUID with UuidRepresentation.UNSPECIFIED. UUIDs can be manually converted to bson.Binary instances using bson.Binary.from_uuid() or a different UuidRepresentation can be configured. See the documentation for UuidRepresentation for more information.

I don't think I tried any further to make that work.

django_mongodb_backend/indexes.py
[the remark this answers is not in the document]
timgraham: Yes, I removed it because it's unused.

django_mongodb_backend/indexes.py
timgraham: I see we can't compute the list length in `__init__()` because we don't know how many vector fields there are, so resolving this thread.

django_mongodb_backend/indexes.py
```diff
@@ -197,20 +195,20 @@ def check(self, model, connection):
                 except (ValueError, TypeError):
                     errors.append(
                         Error(
-                            f"Atlas vector search requires size on {field_name}.",
+                            f"VectorSearchIndex requires 'size' on field '{field_name}'.",
                             obj=model,
-                            id=f"{self._error_id_prefix}.E001",
+                            id=f"{self._error_id_prefix}.E002",
```
WaVEV, line 200: there is no E001, warning count as the first?

django_mongodb_backend/indexes.py
```diff
@@ -164,26 +164,24 @@ def get_pymongo_index_model(
...
-    ALLOWED_SIMILARITY_FUNCTIONS = frozenset(("euclidean", "cosine", "dotProduct"))
+    VALID_SIMILARITIES = frozenset(("euclidean", "cosine", "dotProduct"))
     _error_id_prefix = "django_mongodb_backend.indexes.VectorSearchIndex"
 
     def __init__(self, *expressions, fields=(), similarities="cosine", name=None, **kwargs):
         super().__init__(*expressions, fields=fields, name=name, **kwargs)
         self.similarities = similarities
-        for func in similarities if isinstance(similarities, list) else (similarities,):
-            if func not in self.ALLOWED_SIMILARITY_FUNCTIONS:
+        self._multiple_similarities = isinstance(similarities, tuple | list)
```
WaVEV, line 173: 🤔 multiple_similarities is a local variable.

timgraham: It's also used in `check()`.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+        """
+        Map a model field's internal type to search index type.
+        Reference: https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#data-types
+        """
+        if field.get_internal_type() == "UUIDField":
+            return "uuid"
+        if field.get_internal_type() in {"ObjectIdAutoField", "ObjectIdField"}:
+            return "ObjectId"
+        if field.get_internal_type() == "EmbeddedModelField":
+            return "embeddedDocuments"
```
timgraham, line 138: This is the type we also will want for EmbeddedModelArrayField, correct? We might check if `EmbeddedModelField.db_type()` should be defined and could return `embeddedDocuments`.

How about support for ArrayField? https://www.mongodb.com/docs/atlas/atlas-search/field-types/array-type/

And JSONField? (search type="document"?)

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+        if db_type in {"int", "long"}:
+            return "number"
```
timgraham, line 140: "You can use the Atlas Search number type to index fields with numeric values of int32, int64, and double data types."
So should this include "double" (FloatField) as well?

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+                        "The index won't be created. Use an Atlas-enabled version of MongoDB, "
+                        "or silence this warning if you don't care about it."
+                    ),
+                    obj=model,
+                    id=f"{self._error_id_prefix}.W001",
+                )
+            )
+        return errors
+
+    def search_index_data_types(self, field, db_type):
```
timgraham, line 128: Perhaps the signature could be simplified to `(self, internal_type, db_type)` since `field` isn't used more than `field.get_internal_type()`. Although, see comments below... maybe we don't need `internal_type` after all.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+        for field_name, _ in self.fields_orders:
+            if field_name in seen_fields:
+                raise ValueError(f"Field '{field_name}' is duplicated in fields.")
+            seen_fields.add(field_name)
+
+    def check(self, model, connection):
+        errors = super().check(model, connection)
+        expected_similarities = 0
+        for field_name, _ in self.fields_orders:
+            field_ = model._meta.get_field(field_name)
```
timgraham, line 190: I think the underscore suffix on "field" is an artifact of a previous location of these checks. As far as I see, it's not needed.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+                    int(field_.size)
+                except (ValueError, TypeError):
+                    errors.append(
+                        Error(
+                            f"VectorSearchIndex requires 'size' on field '{field_name}'.",
+                            obj=model,
+                            id=f"{self._error_id_prefix}.E002",
+                        )
+                    )
+                if not isinstance(field_.base_field, FloatField | DecimalField):
```
timgraham, line 203: Is `isinstance()` definitely what we want here as opposed to say checking `db_type()`? (That would be double and decimal for those fields.) On the one hand, an error message that references FloatField & DecimalField probably covers most common use cases and is going to be easier to understand that referencing double/decimal. Just wondering if you had any thoughts on this.

Is this the relevant documentation for this restriction? https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#about-the-vector-type  I see double there but it's not clear to me how DecimalField qualifies. I'm not sure what the doucmentation means when it mentions: 
- BSON BinData vector subtype float32
- BSON BinData vector subtype int1
- BSON BinData vector subtype int8

tests/indexes_/test_search_indexes.py
```diff
@@ -0,0 +1,198 @@
...
+                        "indexDoubles": True,
+                        "indexIntegers": True,
+                        "representation": "double",
+                        "type": "number",
+                    },
+                },
+            }
+            self.assertCountEqual(index_info[index.name]["columns"], index.fields)
+            self.assertEqual(index_info[index.name]["options"], expected_options)
+            self.assertAddRemoveIndex(editor, Article, index)
```
timgraham, line 153: Doesn't this add the index redundantly? `editor.add_index(index=index, model=Article)` is already called above (same in the similar vector search test).

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+                            f"ArrayField '{field_.name}' to be FloatField or "
+                            "DecimalField but is "
+                            f"{field_.base_field.get_internal_type()}.",
+                            obj=model,
+                            id=f"{self._error_id_prefix}.E003",
+                        )
+                    )
+            else:
+                field_type = field_.db_type(connection)
+                search_type = self.search_index_data_types(field_, field_type)
```
timgraham, line 216: The linked doc says, "You can filter on boolean, date, objectId, numeric, string, and UUID values, *including arrays of these types.*"
Does this account for arrays?

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+            if field_name in seen_fields:
+                raise ValueError(f"Field '{field_name}' is duplicated in fields.")
+            seen_fields.add(field_name)
+
+    def check(self, model, connection):
+        errors = super().check(model, connection)
+        expected_similarities = 0
+        for field_name, _ in self.fields_orders:
+            field_ = model._meta.get_field(field_name)
+            if isinstance(field_, ArrayField):
```
timgraham, line 191: I guess a VectorSearchIndex requires at least one ArrayField? Not sure we need a check for that, just wondering.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+        return path, args, kwargs
+
+    def get_pymongo_index_model(
+        self, model, schema_editor, field=None, unique=False, column_prefix=""
+    ):
+        if not schema_editor.connection.features.supports_atlas_search:
+            return None
+        similarities = (
+            itertools.cycle([self.similarities])
+            if isinstance(self.similarities, str)
```
timgraham, line 256: It might be more clear to use `self._multiple_similarities` here.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,176 @@ def where_node_idx(self, compiler, connection):
...
+
+    def search_index_data_types(self, field, db_type):
+        """
+        Map a model field's internal type to search index type.
+        Reference: https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/#data-types
+        """
+        if field.get_internal_type() == "UUIDField":
+            return "uuid"
+        if field.get_internal_type() in {"ObjectIdAutoField", "ObjectIdField"}:
+            return "ObjectId"
```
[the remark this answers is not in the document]
WaVEV: Yes, it should. ObjectId is the BSON Type

django_mongodb_backend/indexes.py
WaVEV: Yes

django_mongodb_backend/indexes.py
WaVEV: JSONField is supported if I am not mistaken. It types as object so is mapped as document. We could map ArrayField also. I don't see any cons to define `EmbeddedModelField.db_type()` as `embeddedDocuments`.

django_mongodb_backend/indexes.py
WaVEV: The only one that would remain is UUIDField. But we could handle it as a string.

tests/indexes_/test_search_indexes.py
WaVEV: Yes, I could be replaced by `editor.remove_index`

django_mongodb_backend/indexes.py
WaVEV: DecimalField is not used for that, is Integer field. But Given that not all the similarities are supported for int1. Shall we support only float32?

django_mongodb_backend/indexes.py
WaVEV: Well, will try to add some light over this behaviours. The array could be defined in the `left hand side` if I try to make a query like:
```
db.t.t.aggregate([
  {
    "$vectorSearch": {
      "index": "vector_index",
      "filter": {
        "name": ["example", "example3"]
      },
      "path": "values",
      "queryVector": [0,0,0,0,0],
      "numCandidates": 150,
      "limit": 10,
      "quantization": "scalar"
    }
  }])
```

I got the following error:
`MongoServerError[UnknownError]: PlanExecutor error during aggregation :: caused by :: "filter" must be a boolean, objectId, number, string, date, uuid, or null`
So I cannot pass an array as a filter.

But I can define a documents like:
```
db.t.t.insertOne({
  name: "example",
  values: [1.23, 4.56, 7.89, 0.12,3.45]
})

db.t.t.insertOne({
  name: ["example", "example2"],
  values: [1.23, 4.56, 7.89, 0.12,3.45]
})
```
Then I filter by:
```
db.t.t.aggregate([
  {
    "$vectorSearch": {
      "index": "vector_index",
      "filter": {
        "name": "example2"
      },
      "path": "values",
      "queryVector": [0,0,0,0,0],
      "numCandidates": 150,
      "limit": 10,
      "quantization": "scalar"
    }
  }])
```
I will get the second and if I filter by example I will get both.

django_mongodb_backend/indexes.py
WaVEV: Nop, It could be created without one.

django_mongodb_backend/indexes.py
WaVEV: But indeed we could support array of those types.

django_mongodb_backend/indexes.py
WaVEV: Supporting ArrayField would be as I explained in [here](https://github.com/mongodb/django-mongodb-backend/pull/264#discussion_r2055182821). It can, will add and if it is not very convincing, we could remove it.

tests/indexes_/test_search_indexes.py
```diff
@@ -146,20 +133,21 @@ def test_multiple_fields(self):
...
 
 @skipUnlessDBFeature("supports_atlas_search")
-class VectorSearchIndexSchemaTests(SchemaAssertionMixin, TestCase):
+class VectorSearchIndexSchemaTests(TestCase):
     def test_simple_vector_search(self):
         with connection.schema_editor() as editor:
             index = VectorSearchIndex(name="recent_article_idx", fields=["number"])
             editor.add_index(index=index, model=Article)
-            self.assertAddRemoveIndex(editor, Article, index)
+            editor.remove_index(index=index, model=Article)
```
timgraham, line 150: But by removing`assertAddRemoveIndex()` you lost the assertion that the index was created.

tests/indexes_/models.py
```diff
@@ -10,6 +10,7 @@ class Data(EmbeddedModel):
 
 class Article(models.Model):
     headline = models.CharField(max_length=100)
+    object_id = ObjectIdField()
```
timgraham, line 13: I think it would be cleaner to create a separate model (`SearchIndexTestModel`?) with all the fields needed for testing SearchIndex data types rather than adding a bunch of non-semantic field to `Article`. As is, it doesn't seem like we have complete test coverage for `search_index_data_types()`.

tests/indexes_/test_search_indexes.py
```diff
@@ -146,20 +133,21 @@ def test_multiple_fields(self):
                         "representation": "double",
                         "type": "number",
                     },
+                    "object_id": {"type": "objectId"},
                 },
             }
             self.assertCountEqual(index_info[index.name]["columns"], index.fields)
             self.assertEqual(index_info[index.name]["options"], expected_options)
-            self.assertAddRemoveIndex(editor, Article, index)
+            editor.remove_index(index=index, model=Article)
```
timgraham, line 141: If you look at `schema/tests.py`, you'll see that this sort of clean up should be in a `finally:` block. You can create multiple `with connection.schema_editor() as editor:` blocks so that the entire method isn't indented under a single one.

django_mongodb_backend/indexes.py
```diff
@@ -125,18 +125,12 @@ def check(self, model, connection):
             )
         return errors
 
-    def search_index_data_types(self, field, db_type):
+    def search_index_data_types(self, db_type):
         """
         Map a model field's internal type to search index type.
```
timgraham, line 130: chop "internal"

tests/indexes_/test_search_indexes.py
WaVEV: Right. haha. Ok, maybe only assertRemoveIndex

django_mongodb_backend/indexes.py
WaVEV: For Vector search I cannot add the support as simple as Search Indexes. It might need something that separates filter fields from vector fields. It will change the class sign. It is not very difficult but I suggest to work on that in a next iteration.

django_mongodb_backend/indexes.py
timgraham: But in practice, is there a use for a vector search index without any vectors? (or have I misunderstood something?)

WaVEV: No, in that case is better to use SearchIndex. MongoDB doesn't restrict that but we can. Will add this check

docs/source/ref/models/indexes.rst
```diff
@@ -0,0 +1,51 @@
...
+.. class:: SearchIndex(fields=(), name=None)
+
+.. versionadded:: 5.2.0b0
+
+Creates a basic :doc:`search index <atlas:atlas-search/index-definitions>` on
+the given field(s).
+
+Some fields such as :class:`~django.db.models.DecimalField` aren't
+supported. See the :ref:`Atlas documentation <atlas:bson-data-chart>` for a
+complete list of unsupported data types.
```
timgraham, line 23: Is this accurate? It might be useful to have a check, but maybe we're spending too much time on this. :wink:

timgraham (commented):
I think we're missing schema tests with `similarities`, both as a string and as a list.

django_mongodb_backend/indexes.py
timgraham: I'm not sure, maybe we ask the team.

docs/source/ref/models/indexes.rst
WaVEV: yes, we are. Most of the time wasted was my fault. 
Yes, it is.

django_mongodb_backend/indexes.py
```diff
@@ -101,6 +105,182 @@ def where_node_idx(self, compiler, connection):
...
+        )
+
+
+class VectorSearchIndex(SearchIndex):
+    suffix = "vsi"
+    VALID_SIMILARITIES = frozenset(("cosine", "dotProduct", "euclidean"))
+    VALID_FIELD_TYPES = frozenset(("boolean", "date", "number", "objectId", "string", "uuid"))
+    _error_id_prefix = "django_mongodb_backend.indexes.VectorSearchIndex"
+
+    def __init__(self, *, fields=(), similarities="cosine", name=None):
```
timgraham, line 170: How did you decide that cosine should be the default?

docs/source/ref/models/indexes.rst
```diff
@@ -0,0 +1,54 @@
...
+.. class:: VectorSearchIndex(fields=(), similarities="cosine", name=None)
+
+.. versionadded:: 5.2.0b0
+
+A subclass of :class:`SearchIndex` that creates a :doc:`vector search index
+<atlas:atlas-vector-search/vector-search-type>` on the given field(s).
+
+The index should reference at least one vector field: an :class:`.ArrayField`
+with a :attr:`~.ArrayField.base_field` of :class:`~django.db.models.FloatField`
+or :class:`~django.db.models.IntegerField`. It cannot reference an
```
timgraham, line 43: Does ArrayField(IntegerField) store data in the correct format for this index? binData(int8)? https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/#about-the-similarity-functions

django_mongodb_backend/indexes.py
WaVEV: To be honest, any time that I worked with embedding, this similarity was the first that I tried. Except when the data was normalized (in that case, cosine and dot product gives the same result and dot product is faster). L2 norm is used but less than cosine in semantics searches.
In order to simplify the index, I decided to put cosine in default.
I have no preference to put similarities as a needed parameter

docs/source/ref/models/indexes.rst
WaVEV: It is stored as array, isn't the best data structure to save this kind of data. But, it works. I remember having this conversation with Jib and James about that.

django_mongodb_backend/indexes.py
timgraham: Reading the Atlas docs, I didn't get the sense that there is a sensible default that works for most situations, but I really don't know, so let's ask the team's opinion on Monday.

timgraham (approved):

PR #288, opened 2025-04-16
PYTHON-5188 Make version setting a part of the release process

blink1073:


blink1073:
Here's a run: https://github.com/mongodb/django-mongodb-backend/actions/runs/14518622648

I had to update the version to a dev version, otherwise it complains that the tag already exists.

Jibola (approved):

PR #328, opened 2025-06-26
Add write concern support for issue #6

scarlet2131:
This PR implements write concern support to replace the broken implementation that didn't work with latest PyMongo APIs.
Changes:

- Add WriteConcern import and parsing in DatabaseWrapper

- Support dict, string, and int configuration formats

- Apply write concern automatically to all collections

- Add comprehensive tests

Usage:

`DATABASES = {
    'default': {
        'ENGINE': 'django_mongodb_backend',
        'NAME': 'mydb',
        'OPTIONS': {
            'WRITE_CONCERN': {'w': 'majority', 'j': True}  # or 'majority' or 2
        }
    }
}`

### DIFF
--- a/django_mongodb_backend/base.py
+++ b/django_mongodb_backend/base.py
@@ -10,6 +10,7 @@ from django.utils.functional import cached_property
 from pymongo.collection import Collection
 from pymongo.driver_info import DriverInfo
 from pymongo.mongo_client import MongoClient
+from pymongo.write_concern import WriteConcern
 
 from . import __version__ as django_mongodb_backend_version
 from . import dbapi as Database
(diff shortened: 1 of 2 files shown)


scarlet2131:
Hey Tim, thanks for looking at this! I think there might be some confusion about what my code actually does.

The crash you mentioned won't happen because I'm not passing `WRITE_CONCERN` to MongoClient. My code pulls it out of OPTIONS first and handles it separately:

```python
# This extracts WRITE_CONCERN before MongoClient sees it
self._write_concern = self._parse_write_concern(settings_dict.get('OPTIONS', {}).get('WRITE_CONCERN'))
```

**So the difference is:**
- Your way: Put `w`, `j`, etc. directly in OPTIONS → MongoClient handles it (client-level)  
- My way: Put them under `WRITE_CONCERN` → I handle it per-collection

I went with collection-level because I thought it gave more control, but honestly I'm not sure which approach you'd prefer for this project. Would client-level write concern work better? I can easily change it if that's what makes more sense.

Just let me know what direction you want to go - happy to adjust!

Jibola:
FYI, I'm closing this PR as it has not received a response in over 3 weeks. Please feel free to reopen if you have additional comments or reopen our issue ticket if you feel it's pertinent for your use case.

Thanks for raising the issue, and always feel free to engage with the repo!

PR #337, opened 2025-07-10
Root doc should be index

aclark4life:
I love the table of contents but it's jarring to navigate to it from the logo text instead of navigating to index.


timgraham (approved):
I agree. I copied root_doc from Django's documentation, but it seems the theme for Django's docs doesn't treat root_doc in the same way (linking to it from a navigation header). Arguably, it should be changed in Django as well, but I'm not sure what it would affect.

PR #381, opened 2025-08-29
INTPYTHON-732 Don't parse the query string args

aclark4life:
Instead, pass them to HOST.

# Not ready for review and potentially targeting post-GA fast follow for this.


aclark4life:
Prefer #383

PR #392, opened 2025-09-08
WIP: Expr support getfield

Jibola:
**TODO**
- Expand queryoptimizer to include getfield support.

### DIFF
--- a/tests/expression_converter_/test_match_conversion.py
+++ b/tests/expression_converter_/test_match_conversion.py
@@ -1,0 +1,216 @@
+from django.test import SimpleTestCase
+
+from django_mongodb_backend.query_conversion.query_optimizer import QueryOptimizer
+
+
+class QueryOptimizerTests(SimpleTestCase):
+    def assertOptimizerEqual(self, input, expected):
+        result = QueryOptimizer().convert_expr_to_match(input)
+        self.assertEqual(result, expected)
+
+    def test_multiple_optimizable_conditions(self):
+        expr = {
+            "$expr": {
+                "$and": [
+                    {"$eq": ["$status", "active"]},
+                    {"$in": ["$category", ["electronics", "books"]]},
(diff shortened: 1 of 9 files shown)


tests/expression_converter_/test_match_conversion.py
```diff
@@ -213,3 +213,69 @@ def test_deeply_nested_logical_operator_with_variable(self):
             }
         ]
         self.assertOptimizerEqual(expr, expected)
+
+    def test_getfield_usage_on_dual_binary_operator(self):
+        expr = {
+            "$expr": {
+                "$gt": [
+                    {"$getField": {"input": "$price", "field": "value"}},
```
NoahStapp, line 221: Can we optimize this down to remove the `$getField` when the field names don't contain `$` or `.`? Something like:

```
{  
  "$match": {  
    "$expr": {  
      "$gt": ["$price.value", "$discounted_price.value"]  
    }  
  }  
}  
```

Jibola: We _could_. I'm trying to think if there's any case where this would be an issue if it mistakenly resolved it.

Jibola: This would have to be a two-phase conversion. I think it may even be best to still change it in our actual query code.

NoahStapp: I'm assuming here that this change ends up being needed. Holding off until we either get customer complaints or the refactor can't solve this makes more sense to me for now.

Jibola: Yeah, I tried to make it work (and I think I managed to), but the silent mutation of `$getField` to `$field` is worrying to me. Even if I manage to make things pass tests, it just feels out of scope of the optimizers goal, which is to explicitly change things from `$expr` to `$match`.

PR #396, opened 2025-09-09
WIP lookup refactor

WaVEV:


django_mongodb_backend/functions.py
```diff
@@ -271,29 +293,35 @@ def trunc_time(self, compiler, connection):
...
-    Now.as_mql = now
-    NullIf.as_mql = null_if
-    Replace.as_mql = replace
-    Round.as_mql = round_
-    RTrim.as_mql = trim("rtrim")
-    StrIndex.as_mql = str_index
-    Substr.as_mql = substr
-    Trim.as_mql = trim("trim")
-    TruncBase.as_mql = trunc
+    Cast.as_mql_expr = cast
```
WaVEV, line 283: All the function does not support `as_mql_path`. It could be added latter if we try to simplify constants expressions

django_mongodb_backend/expressions/builtins.py
```diff
@@ -209,21 +221,37 @@ def value(self, compiler, connection):  # noqa: ARG001
     return value
 
 
+def base_expression(self, compiler, connection, as_path=False, **extra):
```
WaVEV: This is the common handler for all expressions. It defines if an `expr` is needed or not.

django_mongodb_backend/fields/json.py
```diff
@@ -139,37 +177,83 @@ def key_transform_is_null(self, compiler, connection):
...
+    KeyTransform.as_mql_path = partialmethod(key_transform, as_path=True)
+    KeyTransform.as_mql_expr = partialmethod(key_transform, as_path=False)
+    KeyTransformIn.as_mql_path = key_transform_in_path
+    KeyTransformIn.as_mql_expr = key_transform_in_expr
+    KeyTransformIsNull.as_mql_path = key_transform_is_null_path
+    KeyTransformIsNull.as_mql_expr = key_transform_is_null_expr
+    KeyTransformNumericLookupMixin.as_mql_path = key_transform_numeric_lookup_mixin_path
+    KeyTransformNumericLookupMixin.as_mql_expr = key_transform_numeric_lookup_mixin_expr
+    KeyTransformExact.as_mql_expr = key_transform_exact_expr
+    KeyTransformExact.as_mql_path = key_transform_exact_path
```
timgraham: Please check alphabetization of the classes and functions (not only in this file).

django_mongodb_backend/base.py
```diff
@@ -118,19 +123,56 @@ def _isnull_operator(a, b):
...
+        "endswith": lambda a, b: regex_expr(a, (b, {"$literal": "$"})),
+        "iendswith": lambda a, b: regex_expr(a, (b, {"$literal": "$"}), insensitive=True),
+        "contains": lambda a, b: regex_expr(a, b),
+        "icontains": lambda a, b: regex_expr(a, b, insensitive=True),
+        "regex": lambda a, b: regex_expr(a, b),
+        "iregex": lambda a, b: regex_expr(a, b, insensitive=True),
+    }
+
+    def range_match(a, b):
+        ## TODO: MAKE A TEST TO TEST WHEN BOTH ENDS ARE NONE. WHAT SHALL I RETURN?
```
timgraham: AI says, "If either the start or end value provided to the BETWEEN operator is NULL, the entire BETWEEN condition will typically evaluate to UNKNOWN (and thus FALSE in a WHERE clause), unless explicitly handled." (I confirmed this for SQLite and PostgreSQL)

However, to match the semantics implemented here where None is treated as min/max date, I would expect `__range=[None, None]` not to filter any values.

django_mongodb_backend/aggregates.py
```diff
@@ -8,14 +8,7 @@
 MONGO_AGGREGATIONS = {Count: "sum"}
 
 
-def aggregate(
-    self,
-    compiler,
-    connection,
-    operator=None,
-    resolve_inner_expression=False,
-    **extra_context,  # noqa: ARG001
```
timgraham: Is the removal of `extra_context` strictly related to this patch? (Mainly wondering, though perhaps it could be a separate trivial PR).

django_mongodb_backend/aggregates.py
```diff
@@ -64,16 +57,16 @@ def count(self, compiler, connection, resolve_inner_expression=False, **extra_co
...
 
 def register_aggregates():
-    Aggregate.as_mql = aggregate
-    Count.as_mql = count
-    StdDev.as_mql = stddev_variance
-    Variance.as_mql = stddev_variance
+    Aggregate.as_mql_expr = aggregate
+    Count.as_mql_expr = count
+    StdDev.as_mql_expr = stddev_variance
+    Variance.as_mql_expr = stddev_variance
```
timgraham, line 72: I see we have `as_mql_expr()`, `as_mql_path()`, and `as_mql(..., as_path=...)`. If this is the way we keep it, it would be good to explain in the design document which objects (aggregate, func, expression, etc.) get which.

I wonder about renaming `as_mql_expr()` or `as_mql_path()` to `as_mql()` (i.e. treating one of paths as the default). Do you think it would be more or less confusing?

django_mongodb_backend/base.py
```diff
@@ -108,7 +108,12 @@ def _isnull_operator(a, b):
         }
         return is_null if b else {"$not": is_null}
 
-    mongo_operators = {
+    def _isnull_operator_match(a, b):
+        if b:
+            return {"$or": [{a: {"$exists": False}}, {a: None}]}
+        return {"$and": [{a: {"$exists": True}}, {a: {"$ne": None}}]}
+
+    mongo_operators_expr = {
```
timgraham: `mongo_expr_operators` might be a more natural word order.

django_mongodb_backend/base.py
```diff
@@ -118,19 +123,56 @@ def _isnull_operator(a, b):
         "lte": lambda a, b: {
             "$and": [{"$lte": [a, b]}, DatabaseWrapper._isnull_operator(a, False)]
         },
-        "in": lambda a, b: {"$in": [a, b]},
+        "in": lambda a, b: {"$in": (a, b)},
```
timgraham: Is this to match the expected MQL in a test?

django_mongodb_backend/fields/embedded_model.py
```diff
@@ -184,16 +197,22 @@ def get_transform(self, name):
...
-    def as_mql(self, compiler, connection, as_path=False):
+    def as_mql_path(self, compiler, connection):
+        previous = self
+        key_transforms = []
+        while isinstance(previous, KeyTransform):
+            key_transforms.insert(0, previous.key_name)
+            previous = previous.lhs
+        mql = previous.as_mql(compiler, connection, as_path=True)
+        mql_path = ".".join(key_transforms)
+        return f"{mql}.{mql_path}"
```
timgraham: Please check my PRs that touch this area to see if you think they could go in before.

django_mongodb_backend/fields/embedded_model_array.py
```diff
@@ -225,6 +230,7 @@ class EmbeddedModelArrayFieldLessThanOrEqual(
 
 class KeyTransform(Transform):
     field_class_name = "EmbeddedModelArrayField"
+    PREFIX_ITERABLE = "item"
```
timgraham: I suppose this could be a separate trivial PR (it also conflicts my PR).

django_mongodb_backend/fields/json.py
```diff
@@ -8,17 +11,20 @@
...
 
-from ..lookups import builtin_lookup
+from ..lookups import builtin_lookup_expr, builtin_lookup_path
 from ..query_utils import process_lhs, process_rhs
 
 
-def build_json_mql_path(lhs, key_transforms):
+def build_json_mql_path(lhs, key_transforms, as_path=False):
     # Build the MQL path using the collected key transforms.
+    if as_path and lhs:
```
timgraham: Maybe it's less confusing to add `if lhs: .... else: return None` under the `if as_path:` (if that's correct). Otherwise, I'm studying this code wondering what happens to the `as_path=True, lhs=None` case as it falls through to the rest of the function).

django_mongodb_backend/fields/json.py
```diff
@@ -61,24 +72,28 @@ def _has_key_predicate(path, root_column, negated=False):
...
+        paths.append(rhs_json_path.as_mql(compiler, connection, as_path=as_path))
     keys = []
     for path in paths:
-        keys.append(_has_key_predicate(path, lhs))
-    if self.mongo_operator is None:
-        return keys[0]
-    return {self.mongo_operator: keys}
+        keys.append(_has_key_predicate(path, lhs, as_path=as_path))
+
+    return keys[0] if self.mongo_operator is None else {self.mongo_operator: keys}
```
timgraham: I'm studying this wondering why this was converted to a ternary. I think it's unrelated. The old version looks simpler to me.

django_mongodb_backend/fields/json.py
```diff
@@ -104,19 +119,22 @@ def key_transform(self, compiler, connection):
     """
     key_transforms = [self.key_name]
     previous = self.lhs
-    # Collect all key transforms in order.
     while isinstance(previous, KeyTransform):
         key_transforms.insert(0, previous.key_name)
         previous = previous.lhs
-    lhs_mql = previous.as_mql(compiler, connection)
-    return build_json_mql_path(lhs_mql, key_transforms)
+    # Collect all key transforms in order.
```
timgraham: Was the location of the comment incorrect? Is it supposed to be "# Traverse to the root column." as below? I'd advocate for corrections to be made separately.

django_mongodb_backend/fields/json.py
```diff
@@ -139,37 +177,83 @@ def key_transform_is_null(self, compiler, connection):
...
+    rhs_mql = process_rhs(self, compiler, connection)
     return _has_key_predicate(lhs_mql, root_column, negated=rhs_mql)
 
 
-def key_transform_numeric_lookup_mixin(self, compiler, connection):
+def key_transform_numeric_lookup_mixin_path(self, compiler, connection):
+    """
+    Return MQL to check if the field exists (i.e., is not "missing" or "null")
+    and that the field matches the given numeric lookup expression.
+    """
```
timgraham: Perhaps identical docstrings don't need to be repeated. We should decide on which version they should appear (probably the first one, whichever that is.)

django_mongodb_backend/functions.py
```diff
@@ -77,26 +77,27 @@ def cast(self, compiler, connection):
         lhs_mql = {"$convert": {"input": lhs_mql, "to": output_type}}
     if decimal_places := getattr(self.output_field, "decimal_places", None):
         lhs_mql = {"$trunc": [lhs_mql, decimal_places]}
+
```
timgraham: revert

django_mongodb_backend/functions.py
```diff
@@ -105,29 +106,43 @@ def extract(self, compiler, connection):
...
+    if self.function is None:
+        raise NotSupportedError(f"{self} may need an as_mql() method.")
+    operator = MONGO_OPERATORS.get(self.__class__, self.function.lower())
+    if as_path:
+        return {"$expr": {f"${operator}": lhs_mql}}
+    return {f"${operator}": lhs_mql}
+
+
+def func_path(self, compiler, connection):  # noqa: ARG001
+    raise NotSupportedError(f"{self} May need an as_mql_path() method.")
```
timgraham:
```suggestion
    raise NotSupportedError(f"{self} may need an as_mql_path() method.")
```

django_mongodb_backend/query.py
```diff
@@ -11,8 +11,6 @@
 from django.db.models.sql.where import AND, OR, XOR, ExtraWhere, NothingNode, WhereNode
 from pymongo.errors import BulkWriteError, DuplicateKeyError, PyMongoError
 
-from .query_conversion.query_optimizer import convert_expr_to_match
```
timgraham, line 14: You can delete all that code too. :-)

django_mongodb_backend/query_utils.py
```diff
@@ -47,7 +51,41 @@ def process_rhs(node, compiler, connection):
...
-def regex_match(field, regex_vals, insensitive=False):
+def regex_expr(field, regex_vals, insensitive=False):
     regex = {"$concat": regex_vals} if isinstance(regex_vals, tuple) else regex_vals
     options = "i" if insensitive else ""
     return {"$regexMatch": {"input": field, "regex": regex, "options": options}}
+
+
+def regex_match(field, regex, insensitive=False):
+    options = "i" if insensitive else ""
+    # return {"$regexMatch": {"input": field, "regex": regex, "options": options}}
```
timgraham: chop

tests/model_fields_/test_embedded_model.py
```diff
@@ -225,6 +225,15 @@ def test_nested(self):
...
         self.assertCountEqual(Book.objects.filter(author__address__city="NYC"), [obj])
 
+    def test_annotate(self):
+        obj = Book.objects.create(
+            author=Author(name="Shakespeare", age=55, address=Address(city="NYC", state="NY"))
+        )
+        book_from_ny = (
+            Book.objects.annotate(city=F("author__address__city")).filter(city="NYC").first()
+        )
+        self.assertCountEqual(book_from_ny.city, obj.author.address.city)
```
timgraham: If it passes in the current code, it could be added separately.

django_mongodb_backend/aggregates.py
WaVEV: Yes, It was imported from django. I copied the extra_context thing but I realize that the codes never uses those extra_context. So I went to remove it. But I agree, it could be in a separate PR.

django_mongodb_backend/aggregates.py
WaVEV: Yes, that was the idea. I’ll explain it in the docs, and we might also consider renaming some methods. The core concept is:

* Every expression has an `as_mql` method.
* In some cases, it’s simpler to implement `as_mql` directly, so those methods don’t follow the common expression flow.
* For other expressions, `as_mql` is a composite function that delegates to `as_path` or `as_expr` when applied.
* The `base_expression.as_mql` method controls *when* these are called and performs boilerplate checks to prevent nesting an `expr` inside another `expr` (a MongoDB 6 restriction).

In short: every object has `as_mql`. Some also define `as_path` and `as_expr`. The `base_expression` coordinates how these methods are used, except for cases where `as_mql` is defined directly.

tests/model_fields_/test_embedded_model.py
WaVEV: mmh. maybe I have to delete it, I create It to validate something. But the check is contained in others test

tests/model_fields_/test_embedded_model_array.py
```diff
@@ -289,6 +291,13 @@ def test_foreign_field_exact(self):
         qs = Tour.objects.filter(exhibit__sections__number=1)
         self.assertCountEqual(qs, [self.egypt_tour, self.wonders_tour])
 
+    def test_foreign_field_exact_expr(self):
```
WaVEV, line 482: Do I have to do more test like this? just make a query and the check the generated sql?

django_mongodb_backend/base.py
WaVEV: Yes

django_mongodb_backend/fields/json.py
WaVEV: 🤔 I will try to rewrite this function, I think I made a mistake here.

django_mongodb_backend/fields/json.py
WaVEV: yes, moved by accident

django_mongodb_backend/fields/embedded_model.py
WaVEV: It should go in before.

tests/model_fields_/test_embedded_model_array.py
timgraham: Yes, an assertion could be added to existing tests where possible.

django_mongodb_backend/aggregates.py
WaVEV: Doc here: [link](https://docs.google.com/document/d/1Z1JI0PW-KOyVM0q9eSfMxC1eMBHKPmHP2dpqABlRZdg/edit?usp=sharing)

django_mongodb_backend/base.py
WaVEV: Raise FullResultSet is the way to go, and EmptyResultSet if the `low` is bigger than `high`.

django_mongodb_backend/fields/json.py
WaVEV: Fixed

tests/model_fields_/test_embedded_model_array.py
WaVEV: Added

django_mongodb_backend/base.py
```diff
@@ -118,19 +124,64 @@ def _isnull_operator(a, b):
...
+                start = start.to_decimal()
+            if isinstance(end, Decimal128):
+                end = end.to_decimal()
+            if start > end:
+                raise EmptyResultSet
+        if not conditions:
+            raise FullResultSet
+        return {"$and": conditions}
+
+    # match, path, find? don't know which name use.
```
WaVEV: I mean, mongo_??_operators

django_mongodb_backend/base.py
```diff
@@ -118,19 +124,64 @@ def _isnull_operator(a, b):
...
+    def range_match(a, b):
+        conditions = []
+        start, end = b
+        if start is not None:
+            conditions.append({a: {"$gte": b[0]}})
+        if end is not None:
+            conditions.append({a: {"$lte": b[1]}})
+        if start is not None and end is not None:
+            if isinstance(start, Decimal128):
+                start = start.to_decimal()
```
WaVEV, line 117: potencial overflow risk here?

django_mongodb_backend/base.py
```diff
@@ -108,7 +109,12 @@ def _isnull_operator(a, b):
         }
         return is_null if b else {"$not": is_null}
 
-    mongo_operators = {
+    def _isnull_operator_match(a, b):
+        if b:
+            return {"$or": [{a: {"$exists": False}}, {a: None}]}
```
Jibola: What does b represent in this case? Not negating?

django_mongodb_backend/base.py
Jibola: `mongo_match_operators` does sound good. It still needs a comment have them briefly explained

django_mongodb_backend/aggregates.py
Jibola: @timgraham I actually like the decoupling of `as_mql` and `as_mql_(path | expr)`. I view it as, you need to define at least two:
`as_mql` and `as_mql_expr`. 

Then if you want to have the more optimized function you define `as_mql_path`. It feels less confusing to me that way.

django_mongodb_backend/base.py
timgraham: Is there a failing test without the `to_decimal()` calls? I'm not sure why there would be a mix of Decimal/Decimal128 here.

django_mongodb_backend/base.py
timgraham: If we consider that "match" is the default path unless expr is needed, I see an argument from simply calling it `mongo_operators` and leaving `mongo_expr_operators` as the name that needs some distinction.

Similarly, the `as_path=False` default argument might become `as_expr=False`. (Would that result in more or fewer  cases to be handled with  `as_expr=True`?

django_mongodb_backend/base.py
Jibola: This is a BSON type held at the database and returned locally. I think it should be good.

django_mongodb_backend/expressions/search.py
```diff
@@ -933,11 +933,16 @@ def __str__(self):
...
+    def as_mql_expr(self, compiler, connection):
+        lhs_mql = process_lhs(self, compiler, connection, as_path=False)
+        value = process_rhs(self, compiler, connection, as_path=False)
         return {"$gte": [lhs_mql, value]}
 
+    def as_mql_path(self, compiler, connection):
+        lhs_mql = process_lhs(self, compiler, connection, as_path=True)
+        value = process_rhs(self, compiler, connection, as_path=True)
+        return {lhs_mql: {"$gte": value}}
+
```
Jibola, line 945: Why is there a `$gte` query in a `search.text` lookup?

django_mongodb_backend/fields/embedded_model_array.py
```diff
@@ -235,13 +240,27 @@ def __init__(self, field, *args, **kwargs):
...
     def __call__(self, this, *args, **kwargs):
         self._lhs = self._sub_transform(self._lhs, *args, **kwargs)
         return self
 
+    @property
+    def can_use_path(self):
+        return self.is_simple_column
+
+    @cached_property
+    def is_simple_column(self):
```
Jibola: I see this validation for embedded models used in multiple places. Can you consolidate this to the query_utils file?

django_mongodb_backend/compiler.py
```diff
@@ -481,11 +481,11 @@ def build_query(self, columns=None):
             query.lookup_pipeline = self.get_lookup_pipeline()
             where = self.get_where()
             try:
-                expr = where.as_mql(self, self.connection) if where else {}
+                match = where.as_mql(self, self.connection, as_path=True) if where else {}
```
Jibola: Since we enforce `as_path=True` here, can we set that as the default on `WhereNode.as_mql(..., as_path=True)`?

django_mongodb_backend/query_utils.py
```diff
@@ -1,41 +1,47 @@
...
+from django.db.models.expressions import CombinedExpression, Func, Value
+from django.db.models.sql.query import Query
 
 
 def is_direct_value(node):
     return not hasattr(node, "as_sql")
 
 
-def process_lhs(node, compiler, connection):
+def process_lhs(node, compiler, connection, as_path=False):
```
Jibola: Based on the delegation logic, should `as_path` always default to `True`?
> If the caller requests a path filter (as_path=True) and the expression supports it, as_mql_path is executed.
If the expression cannot be represented as a path filter but as_path=True, it is automatically encapsulated in $expr.
If the caller requests an expression (as_path=False), as_mql_expr is executed directly. No additional encapsulation is needed, since the query generation ensures proper nesting of $expr.

django_mongodb_backend/query_utils.py
```diff
@@ -47,7 +53,47 @@ def process_rhs(node, compiler, connection):
...
+
+def regex_match(field, regex, insensitive=False):
+    options = "i" if insensitive else ""
+    return {field: {"$regex": regex, "$options": options}}
+
+
+def is_constant_value(value):
+    if isinstance(value, CombinedExpression):
+        # Temporary: treat all CombinedExpressions as non-constant until
+        # constant cases are handled
```
Jibola: Can you make a follow-up ticket for this? If it already exists, could you link it in the comment?

django_mongodb_backend/base.py
WaVEV: yes, there is: `model_fields_.test_polymorphic_embedded_model.QueryingTests.test_range`. Maybe this case shouldn't be handled here, and let the query optimizer to remove those `False` cases.

django_mongodb_backend/query_utils.py
```diff
@@ -47,7 +53,47 @@ def process_rhs(node, compiler, connection):
...
+        isinstance(value, Value)
+        or
+        # Some closed functions cannot yet be converted to constant values.
+        # Allow Func with can_use_path as a temporary exception.
+        (isinstance(value, Func) and value.can_use_path)
+    )
+
+
+def valid_path_key_name(key_name):
+    return bool(re.fullmatch(r"[A-Za-z0-9_]+", key_name))
```
Jibola: https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/

Values like hashtags are also valid for path names and don't require `$expr`. To my knowledge so long as it's not (.) or ($) it's good.

Jibola (changes requested):
Left my first round of comments. I'm going to review the tests as a second phase of the PR.

django_mongodb_backend/base.py
WaVEV: 🤔  well, it is a bit confusing here. but this function checks for nullability. So, if b is True then we check null. maybe the parameter should be: field, null.

django_mongodb_backend/base.py
WaVEV: 🤔 yes, as_expr=False seems more consistent.

django_mongodb_backend/aggregates.py
WaVEV: 🤔  An expression needs at least one of the methods, as_mql_expr is enough (not optimal but functional). The aggregation module is a good example. Because the as_mql is defined in baseExpression class. In this approach **most of the expressions don't need as_mql** (if as_mql is defined, expr or path aren't needed). 
EDIT: Sorry, I rushed the answer before read all the text, you got the point well. 😬

django_mongodb_backend/expressions/search.py
WaVEV: To convert a score function into a filter I decided to express the following proposition: `score_func(...) > 0`.

django_mongodb_backend/fields/embedded_model_array.py
WaVEV: 🤔 will try. The overall structure of the function is similar, but the type varies. I could create a meta-function that generates the appropriate function for a given type.

django_mongodb_backend/compiler.py
WaVEV: Yes, will refactor. Also it may change the name, as_expr=False. but it is scenically the same

django_mongodb_backend/query_utils.py
WaVEV: Yes, I will.

django_mongodb_backend/query_utils.py
WaVEV: Thanks, will adjust the expression.

django_mongodb_backend/fields/embedded_model.py
```diff
@@ -185,21 +200,27 @@ def get_transform(self, name):
...
-        mql = previous.as_mql(compiler, connection)
-        for column in columns:
-            mql = {"$getField": {"input": mql, "field": column}}
+        return columns, previous
+
+    def as_mql_expr(self, compiler, connection):
+        columns, parent_field = self._get_target_path()
+        mql = parent_field.as_mql(compiler, connection)
+        for key in columns:
+            mql = {"$getField": {"input": mql, "field": key}}
```
Jibola: potentially out of scope:
https://github.com/mongodb/django-mongodb-backend/pull/392/files#diff-0a6ce30a131a00fa88086c4c4d0d6e6232845fd11ef2bc67891fdf92e10c3743R18-R45

Is it possible to still remove `$getField` in `as_mql_expr` or is it expected that routing to `as_mql_expr` for embedded model queries is because of needing a `getField` call?

django_mongodb_backend/lookups.py
```diff
@@ -121,13 +138,25 @@ def uuid_text_mixin(self, compiler, connection):  # noqa: ARG001
     raise NotSupportedError("Pattern lookups on UUIDField are not supported.")
 
 
+@property
+def can_use_path(self):
+    simple_column = getattr(self.lhs, "is_simple_column", False)
+    constant_value = is_constant_value(self.rhs)
+    return simple_column and constant_value
```
Jibola: 💯

django_mongodb_backend/base.py
Jibola: field and is_null are good clarifiers

tests/model_fields_/test_arrayfield.py
```diff
@@ -291,24 +337,140 @@ def test_exact_nested(self):
...
+        with self.assertNumQueries(1) as ctx:
+            self.assertSequenceEqual(
+                NullableIntegerArrayModel.objects.filter(field__in=[[1], [2]]),
+                self.objs[:2],
+            )
+        query = ctx.captured_queries[0]["sql"]
+        self.assertAggregateQuery(
+            query,
+            "model_fields__nullableintegerarraymodel",
+            [{"$match": {"field": {"$in": ([1], [2])}}}],
```
Jibola: Does $in now expect a tuple?

tests/model_fields_/test_embedded_model.py
```diff
@@ -144,23 +146,360 @@ def setUpTestData(cls):
...
+        query = ctx.captured_queries[0]["sql"]
+        self.assertAggregateQuery(
+            query,
+            "model_fields__holder",
+            [
+                {
+                    "$match": {
+                        "$expr": {
+                            "$eq": [
+                                {"$getField": {"input": "$data", "field": "integer_"}},
```
Jibola, line 169: This is an example of a value that could get rid of the getField.

tests/model_fields_/test_embedded_model.py
```diff
@@ -144,23 +146,360 @@ def setUpTestData(cls):
...
+                Holder.objects.filter(data__integer__lt=Value(4) - 1), self.objs[:3]
+            )
+        query = ctx.captured_queries[0]["sql"]
+        self.assertAggregateQuery(
+            query,
+            "model_fields__holder",
+            [
+                {
+                    "$match": {
+                        "$expr": {
```
Jibola, line 212: Curious to the callback chain on this one since the null-check could actually be converted. 

May best be an improvement added later

django_mongodb_backend/base.py
WaVEV: Done

WaVEV: Done

django_mongodb_backend/fields/embedded_model_array.py
WaVEV: I didn't get better. 😬 . I will let as it is. there is some details like:
`previous._field.column`, `previous.key_name` when extract the field_name or the path. Then are different from EMFA and EMF the return, one has to validate the inner transform while the other doesn't

tests/model_fields_/test_embedded_model.py
WaVEV: Yes, we could get rid of those getField if we use `$data.integer_` but I thought it was out of scope for this refactor. This behavior is the current behavior.

tests/model_fields_/test_arrayfield.py
WaVEV: Yes, but it wasn’t a new convention or expectation. There was already a [test](https://github.com/mongodb/django-mongodb-backend/blob/main/tests/lookup_/tests.py#L61-L68) that checks the RHS `$in` as a tuple, so I just followed that convention. It doesn’t affect the query behavior.

tests/model_fields_/test_embedded_model.py
Jibola: Yes, it's out of scope!

django_mongodb_backend/query_utils.py
WaVEV: https://jira.mongodb.org/browse/INTPYTHON-783

django_mongodb_backend/query_utils.py
WaVEV: what if there is some emoji or some non ascii character? 🤔

django_mongodb_backend/fields/embedded_model.py
```diff
@@ -185,21 +200,27 @@ def get_transform(self, name):
...
-            return f"{mql}.{mql_path}"
-        mql = previous.as_mql(compiler, connection)
-        for column in columns:
-            mql = {"$getField": {"input": mql, "field": column}}
+        return columns, previous
+
+    def as_mql_expr(self, compiler, connection):
+        columns, parent_field = self._get_target_path()
+        mql = parent_field.as_mql(compiler, connection, as_expr=True)
+        for key in columns:
```
timgraham: Did you mean to revert "column" to "key" (after 6d5537ce37ffb1a5290853f32d0321beb2ac4061)?

tests/model_fields_/test_arrayfield.py
```diff
@@ -241,9 +243,34 @@ def test_empty_list(self):
...
+
+    def test_exact_path(self):
+        with self.assertNumQueries(1) as ctx:
+            self.assertSequenceEqual(
+                NullableIntegerArrayModel.objects.filter(field__exact=[1]), self.objs[:1]
+            )
+        query = ctx.captured_queries[0]["sql"]
+        self.assertAggregateQuery(
+            query, "model_fields__nullableintegerarraymodel", [{"$match": {"field": [1]}}]
         )
```
timgraham: A small nit: For all of these tests, I would order/name them with this convention: `test_exact` (path case), `test_extact_expr`. In other words, the simple case should be first and it doesn't need `_path` suffix.

tests/model_fields_/test_embedded_model_array.py
```diff
@@ -177,23 +179,171 @@ def setUpTestData(cls):
...
-            [self.lost_empires],
+    def test_array_index_expr(self):
+        with self.assertNumQueries(1) as ctx:
+            self.assertCountEqual(
+                Exhibit.objects.filter(sections__0__number=Value(2) - 1),
+                [self.egypt, self.wonders],
+            )
+        query = ctx.captured_queries[0]["sql"]
+        self.assertAggregateQuery(
+            query,
```
timgraham: In existing tests, I omitted the `query` variable, e.g. 
```
self.assertAggregateQuery(
    ctx.captured_queries[0]["sql"],
```

django_mongodb_backend/base.py
timgraham: The root issue (which could be clarified by a comment) is that Decimal128s aren't comparable using less than / greater than:
```
>>> Decimal128("3") < Decimal128("2")
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'Decimal128' and 'Decimal128'
```
Perhaps this is a reasonable feature request that would allow obsoleting this logic in the future?

django_mongodb_backend/fields/embedded_model.py
WaVEV: mmh maybe I forgot to rename it.  Will rename

django_mongodb_backend/fields/embedded_model.py
```diff
@@ -166,6 +168,19 @@ def __init__(self, field, *args, **kwargs):
...
+    @property
+    def can_use_path(self):
+        return self.is_simple_column
+
+    @cached_property
+    def is_simple_column(self):
+        previous = self
+        while isinstance(previous, EmbeddedModelTransform):
+            if not valid_path_key_name(previous._field.column):
+                return False
```
timgraham: The purpose of this could be described with a comment. What's an example of an invalid path key name? And valid_path_key_name could use a comment or docstring. I guess "user.address.city" is a path and "user", "address", etc. are considered "keys"?

WaVEV: A json field could have some rare string as key fields. This [test](https://github.com/mongodb-forks/django/blob/mongodb-5.2.x/tests/model_fields/test_jsonfield.py#L839) shows the case. Yes I called key the things between dots. 🤔 don't know if there is a better name.

timgraham: But this is EmbeddedModelTransform... not JSONField's KeyTranform. Is this tested? Tried `model_fields_` but all passed with `assert False` above the `return False`.

WaVEV: 🤔 I thought I added a test. I forgot to add, just we could define a column like:
`wifi_column = models.IntegerField(db_column="$$Wifi")`.
And now I am thinking of the other models and lookups. will they work If the user start to define columns like that?

But maybe this analysis is out of scope of this ticket 😬

timgraham: Without tests, I'm not sure if there could be other issues. We should ask the team if we want to support it. It seems unlikely based on the discouragements described at https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/#field-names-with-periods-and-dollar-signs.

WaVEV: Test added: `test_query_price_column`. (should be polished but it shows the problem)

django_mongodb_backend/query_utils.py
Jibola: Then how about we change the boolean to be something like this?
`return not bool(re.search(r"[\$\.]")`

tests/lookup_/tests.py
```diff
@@ -1,23 +1,38 @@
...
-        Number.objects.create()
+        cls.null_number = Number.objects.create()
 
     def test_lt(self):
         self.assertQuerySetEqual(Number.objects.filter(num__lt=3), self.objs[:3])
 
     def test_lte(self):
         self.assertQuerySetEqual(Number.objects.filter(num__lte=3), self.objs[:4])
 
+    def test_empty_range(self):
```
WaVEV, line 22: Writing the new batery of test, I discovered that this should be moved inside lookupMQLTest. 😬

PR #407, opened 2025-09-19
Fix EmbeddedModelField crash when retrieving model where field name isn't present in db data

ddrondo:
If the field is embedded and empty, it gave an error about the absence of this field.

### DIFF
--- a/django_mongodb_backend/operations.py
+++ b/django_mongodb_backend/operations.py
@@ -184,6 +184,8 @@ class DatabaseOperations(GISOperations, BaseDatabaseOperations):
         if value is not None:
             # Apply database converters to each field of the embedded model.
             for field in expression.output_field.embedded_model._meta.fields:
+                if field.blank:
+                    continue
                 field_expr = Expression(output_field=field)
                 converters = connection.ops.get_db_converters(
                     field_expr


ddrondo:
```
 "route_fares": {
    "tent": {
    },
  },
```
Не все документы содержат фиксированный формат, иногда нет необходимости указывать поле. Django не умеет это поддерживать, т.к. создана для реляционных бд. Данной реализацией можно частично закрыть необходимость в этом. Я согласен, что blank не подходит под данную задумку. Возможно для этого решения необходимо придумать другой параметр, например, void=True

timgraham:
> Not all documents contain a fixed format, sometimes there is no need to specify a field. Django does not know how to support this, because it was created for relational databases. 

Got it. We've been considering whether or not to try to support this sort of sparse data that wasn't written by Django. We've identified some other problems in passing (#275, #401) but it's unclear that supporting this use case is worth it. Writing rigorous tests for sparse data would be a large effort.

timgraham:
All the data that Django writes expects [converters](https://github.com/mongodb/django-mongodb-backend/blob/e747973a6f49509c2375d05aa062e6aa3920dcbf/django_mongodb_backend/operations.py#L146-L223) to run. If converters don't run, field data will be in an unexpected time (e.g. DateField values will be datetime rather than date, DecimalField will be Decimal128 rather than Decimal). This is why your proposed solution of checking `blank` is incorrect: converters won't run for fields with `blank=True`, even if the field has data.

ddrondo:
I agree with blank=True that it is incorrect to specify it, but your decision is also incorrect. I think the best solution would be to inherit from 'from django.db.models.fields import Field' and add a new attribute to solve the problem

timgraham:
As I explained, by disabling an embedded model field's converters, you're going to break those fields. Instead, for example, if you want a version of DecimalField that uses Decimal128 instead of Decimal, you'll instead want to write a custom field (Decimal128Field) and use that on your embedded model.

Nonetheless, you can implement your own custom embedded model field if you believe your proposed solution is suitable for your project.

ddrondo:
I'm not talking about converters, I'm talking about
```
if field.attname in value:
      value[field.attname] = converter(value[field.attname], field_expr, connection)
```
Not everyone expects this behavior, and many follow a strict pattern of the model.

timgraham:
Please check if https://github.com/mongodb/django-mongodb-backend/pull/409 solves your issue.

If not, you'll need to explain a bit more. I'm not sure what a "strict schema of the model" means.

The lines you quoted is where database converters are run on each embedded model field. These cannot be disabled if a field has any data, otherwise the data won't be converted as required (e.g. an EmbeddedModel's DecimalField value would be loaded as Decimal128 instead of Decimal).

timgraham:
(The conversation can continue if need be, but the problem should addressed by https://github.com/mongodb/django-mongodb-backend/pull/409.)

PR #410, opened 2025-09-25
INTPYTHON-769 Made embedded fields respect Field.db_column

timgraham:


django_mongodb_backend/fields/embedded_model.py
WaVEV, line 209: 🤔 maybe te key_name is not longer needed.

timgraham: I separated the cleanup: #411

timgraham: Also, the first commit is separate: https://github.com/mongodb/django-mongodb-backend/pull/409

PR #414, opened 2025-10-02
INTPYTHON-778 Correct Polymorphic/EmbeddedModelField search index type

timgraham:
"embeddedDocuments" isn't a BSON type.


timgraham:
@WaVEV I guess this is a user facing bug fix since "document" should be used for embedded model search indexes, while [embeddedDocuments is only for arrays of documents](https://www.mongodb.com/docs/atlas/atlas-search/field-types/document-type/?deployment-type=atlas&interface=driver&language=nodejs#document-type-limitations-10). Correct? If so, I will create a ticket and add a release note.

WaVEV (approved):

PR #415, opened 2025-10-04
INTPYTHON-777 Convert multi-level relative imports to absolute

lyova24:
Replaced all relative imports (e.g., `from ...utils import ...`) with in-package absolute imports (e.g., `from django_mongodb_backend.utils import ...`) across the codebase in ./django_mongodb_backend.

[A similar PR for Django](https://github.com/django/django/pull/19707)

FYI: used my own pre-commit-hook [shrimport](https://github.com/lyova24/shrimport)


timgraham (approved):
For future reference... similar to Django, I wouldn't require a Jira issue for cosmetic clean ups.

PR #521, opened 2026-04-16
INTPYTHON-947 Add support for Django’s warning when performing import-time queries

timgraham:
Implements https://github.com/django/django/commit/fbd16438f46bc2128926958ad24331da5d1b406f for this backend.


aclark4life (approved):
Interesting! Thanks

PR #524, opened 2026-04-23
INTPYTHON-950 Re-raise DatabaseError from atomic() when commit fails (opt-in)

Logosph:
Re-raise DatabaseError from atomic() when commit fails (opt-in)

## Changes in this PR

Atomic transactions are used in our production project everywhere. Sometimes there is a situation when MongoDB returns an error when committing a transaction. In this case, the transaction is quietly rolled back. I believe that this behavior is incorrect, and such errors should be thrown further, so that the developer can handle such an exception on their own. Currently, in our project, we have patched `Atomic.__exit__`, but in the future, we would like to get rid of this hack.

What is done:
- Added `raise_on_commit_failure` flag with default value `False` to `atomic` context manager and decorator
- Added logic: 
- - if `raise_on_commit_failure` is set to False, then the behavior remains the same as it was before. This allows you to maintain compatibility with previous versions and allows developers to safely implement a new version of the library.
- - if `raise_on_commit_failure` is set to True, then if MongoDB returns `DatabaseError` on commit transaction, transaction rollbacks and the exception throws further

## Test Plan

Added unit tests with test cases:
1. Using context manager, MongoDB raises `DatabaseError`, `raise_on_commit_failure` is set to True, exception was raised and new record in collection didn't appear
2. Using decorator, MongoDB raises `DatabaseError`, `raise_on_commit_failure` is set to True, exception was raised and new record in collection didn't appear
3. Using context manager, MongoDB raises `DatabaseError`, `raise_on_commit_failure` is set to False, exception was not raised and new record in collection didn't appear
4. Using decorator, MongoDB raises `DatabaseError`, `raise_on_commit_failure` is set to False, exception was not raised and new record in collection didn't appear
5. Using context manager, MongoDB raises `DatabaseError`, `raise_on_commit_failure` isn't set, exception was not raised and new record in collection didn't appear
4. Using decorator, MongoDB raises `DatabaseError`, `raise_on_commit_failure` isn't set, exception was not raised and new record in collection didn't appear

### Checklist for Author

- [x] Did you update the changelog (if necessary)? - I didn't find it
- [x] Is the intention of the code captured in relevant tests?
- [x] If there are new TODOs, has a related JIRA ticket been created?

### Checklist for Reviewer

### DIFF
--- a/django_mongodb_backend/transaction.py
+++ b/django_mongodb_backend/transaction.py
@@ -16,8 +16,9 @@ class Atomic(ContextDecorator):
     Simplified from django.db.transaction.
     """
 
-    def __init__(self, using):
+    def __init__(self, using, raise_on_commit_failure):
         self.using = using
+        self.raise_on_commit_failure = raise_on_commit_failure
 
     def __enter__(self):
         connection = get_connection(self.using)
@@ -45,6 +46,8 @@ class Atomic(ContextDecorator):
                     connection.commit_mongo()
                 except DatabaseError:
                     connection.rollback_mongo()
+                    if self.raise_on_commit_failure:
+                        raise
         else:
             # atomic() exited with an error.
             if not connection.in_atomic_block_mongo:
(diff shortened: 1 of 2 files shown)


django_mongodb_backend/transaction.py
```diff
@@ -45,17 +46,19 @@ def __exit__(self, exc_type, exc_value, traceback):
                     connection.commit_mongo()
                 except DatabaseError:
                     connection.rollback_mongo()
+                    if self.raise_on_commit_failure:
+                        raise
```
timgraham, line 50: Thanks for the report. In fact, I think it's a bug because it's inconsistent with Django's `transaction.atomic()`. See https://github.com/django/django/blob/d24160ab4855f6971aa1f437720adca6330048a2/django/db/transaction.py#L261-L271. This should be simply `raise`. I don't think there's a need for `raise_on_commit_failure`.

I think https://github.com/mongodb/django-mongodb-backend/pull/525 should be sufficient. Please check.

timgraham: Closing in favor of https://github.com/mongodb/django-mongodb-backend/pull/525.

PR #534, opened 2026-05-18
INTPYTHON-530: Run performance tests on a regular cadence

sophiayangDB:
[INTPYTHON-530](https://jira.mongodb.org/browse/INTPYTHON-530)

## Changes in this PR

Modification to the Evergreen config to support regular performance testing. Prior to this change, the performance tests were only run commits a maximum of once every 24 hours. After this change, performance tests should be run daily with or without any new commits.

## Test Plan

Check on Evergreen that the performance tests are running at a daily cadence instead of once a day only with commits.

### Checklist for Author

- [n/a] Did you update the changelog (if necessary)?
- [n/a] Is the intention of the code captured in relevant tests?
- [n/a] If there are new TODOs, has a related JIRA ticket been created?

### Checklist for Reviewer

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -247,6 +247,6 @@ buildvariants:
     display_name: Performance Tests
     run_on:
       - rhel90-dbx-perf-large
-    cron: '0 16 * * MON'  # Daily
+    cron: '0 16 * * *'  # Run daily at 16:00 UTC
     tasks:
       - name: run-performance-tests
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -247,6 +247,6 @@ buildvariants:
     display_name: Performance Tests
     run_on:
       - rhel90-dbx-perf-large
-    batchtime: 1440  # Daily
+    cron: '0 16 * * MON'  # Daily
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -247,6 +247,6 @@ buildvariants:
     display_name: Performance Tests
     run_on:
       - rhel90-dbx-perf-large
-    batchtime: 1440  # Daily
+    cron: '0 16 * * MON'  # Daily
```
sleepyStick: because i never remember at cron formatting -- mind updating this comment to describe the frequency? like daily at what time? 
(but also i think the MON is referring to Monday?? so just general confusion about the comment here lol)

sleepyStick (changes requested):

.evergreen/config.yml
sophiayangDB: Copy and pasting was not the smartest move lol

sleepyStick: HAHAH happens to us all

sophiayangDB:
Decision was made to only run daily if there are any new commits

PR #540, opened 2026-05-27
Disable performance tests on pull requests

timgraham:
This approach isn't ideal because it won't allow the tests to be triggered on a PR manually (for example, when new benchmarks are added). Instead, we might follow the "pr" tag approach that the mongodb-python-driver Evergreen config uses.

However, I think we should revisit this later if that solution seems justified. Temporarily reverting this patch in a PR branch may work to check the performance tests there if need be. Evergreen shows we waste $0.30 each time this job runs on a PR since the results aren't used. For comparison, running the non-performance tests costs about $0.20 total.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -173,6 +173,7 @@ tasks:
       - func: "teardown csfle"
 
   - name: run-performance-tests
+    patchable: false
     commands:
       - func: "run performance tests"
       - func: "attach performance test results"


timgraham:
This approach isn't correct because it won't allow the tests to be triggered on a PR manually (for example, when new performance benchmarks are added). Instead, we might follow the "pr" tag approach that the mongodb-python-driver evergreen config uses.

timgraham:
I think we should merge this and revisit it later if a more elegant solution seems justified. I've updated the commit message and PR description with my rationale.

aclark4life (approved):

PR #543, opened 2026-05-28
INTPYTHON-855 Create unique constraints in a way that can be used by the query planner

sophiayangDB:
INTPYTHON-855

## Changes in this PR
Before this change, exact lookups were generated as plain equality predicates in lookups.py. Since those queries did not include the type conditions required by partial unique indexes, MongoDB often chose a collection scan instead of using the index. This patch adds the corresponding type filters to exact lookups so the planner can select the intended partial unique index.

## Test Plan
Added unit tests in `tests/constraints_`

### Checklist for Author

- [x] Did you update the changelog (if necessary)?
- [x] Is the intention of the code captured in relevant tests?
- [x] If there are new TODOs, has a related JIRA ticket been created?

### Checklist for Reviewer

- [x] Does the title of the PR reference a JIRA Ticket?
- [x] Do you fully understand the implementation? (Would you be comfortable explaining how this code works to someone else?)
- [x] Have you checked for spelling & grammar errors?
- [x] Is all relevant documentation (README or docstring) updated?


django_mongodb_backend/lookups.py
```diff
@@ -26,6 +27,32 @@ def builtin_lookup_path(self, compiler, connection):
...
+
+    return {lhs_mql: {"$type": db_type}}
+
+
+def exact_path(self, compiler, connection):
+    query = builtin_lookup_path(self, compiler, connection)
+    type_filter = _type_filter_for_partial_unique_index(self, compiler, connection)
+    if type_filter is None:
+        return query
+    return {"$and": [query, type_filter]}
```
timgraham: I believe the test failures are because when a type filter is added here, it prevents any null values from being returned. Did you miss the solution suggested on the Jira ticket of creating the indexes in a different way?

django_mongodb_backend/constraints.py
```diff
@@ -9,6 +10,41 @@
...
+            }
+        case "bool":
+            return {"$in": [True, False]}
+        case "date":
+            return {
+                "$gte": datetime.datetime.min,
+                "$lte": datetime.datetime.max,
+            }
+        case _:
+            return {"$exists": True}
```
timgraham: `$exists` matches if the field is null, which we don't want (that explains at least some of the test failures).

Consider this advice: "Null vs. Missing: $exists: true will return documents where a field is explicitly set to null. If you want documents where a field exists and is not null, combine it with $ne: null"

django_mongodb_backend/constraints.py
```diff
@@ -9,6 +10,41 @@
 from .fields import EmbeddedModelArrayField, PolymorphicEmbeddedModelArrayField
 from .indexes import EmbeddedFieldIndexMixin, _get_condition_mql, get_field
 
+def _get_partial_unique_filter(field, connection):
+    db_type = field.db_type(connection)
+    if db_type is None:
+        return {"$exists": True}
```
timgraham: Is this branch used? Offhand, I can't think of why `db_type` would be `None`. That would denote a virtual field that don't exist in the database, and such fields wouldn't be indexed... (perhaps this is AI nonsense.)

django_mongodb_backend/constraints.py
```diff
@@ -10,6 +11,38 @@
...
+                "$lte": datetime.datetime.max,
+            }
+        case _:
+            return {"$type": db_type}
+
+
+def _add_nullable_unique_filter(filter_expression, field_or_field_column, connection, column):
+    field = getattr(field_or_field_column, "field", field_or_field_column)
+    if not field.null:
+        return
```
timgraham: Conceptually, it makes sense to me that the partialFilterExpression isn't needed if the field can't contain null values. My concern is that we would also have to make AlterField modify any/drop any relevant indexes when a field changes from null to not-null (or vice versa).

blink1073:
I'd recommend:

- adding "double" and "decimal" cases to `_get_partial_unique_filter`
- adding unit test coverage for the remaining filter shapes
- adding a changelog entry about the bug fix

timgraham:
Extending what Steve said:
- In docs/ref/models/constraints.rst, we need to document which unique constraints will be used by the query planner and which won't (anything that uses $type, I believe?).
- Adding more details about the changelog entry: we need a mention in docs/releases/6.0.x.rst under "Backwards incompatible changes" noting that users will need to drop and re-create indexes if they want their indexes to adopt the new behavior.

django_mongodb_backend/constraints.py
```diff
@@ -10,6 +11,44 @@
...
+        case "string":
+            return {"$gte": ""}  # Matches all strings including empty string
+        case "int":
+            return {
+                "$gte": -2147483648,  # Min 32-bit integer
+                "$lte": 2147483647,  # Max 32-bit integer
+            }
+        case "long":
+            return {
+                "$gte": -9223372036854775808,  # Min 64-bit integer
```
blink1073: Double and long should be using the 64bit floating point range, not integer range

blink1073 (changes requested):

django_mongodb_backend/constraints.py
timgraham: Are you sure about a float range for long? They are stored as `Int64` as of 4247cead121d92cbe9e57f8b14f9d319fbd1f1de.

The [docs for BSON type](https://www.mongodb.com/docs/manual/reference/bson-types/) "long" says "64-bit integer".

blink1073: Ah, you are right, just double needs to be adjusted

blink1073 (approved):
LGTM

django_mongodb_backend/constraints.py
```diff
@@ -10,6 +13,44 @@
...
+            return {
+                "$gte": -sys.float_info.max,
+                "$lte": sys.float_info.max,
+            }
+        case "decimal":
+            return {
+                "$gte": Decimal128("-9999999999999999999999999999999999E6111"),
+                "$lte": Decimal128("9999999999999999999999999999999999E6111"),
+            }
+        case "bool":  # For consistency
```
timgraham: I'm not sure what the comment "For consistency" means?

docs/ref/models/constraints.rst
```diff
@@ -23,6 +23,14 @@ If you wish to only allow one document with a ``NULL`` value, use a
...
     <django.db.models.UniqueConstraint.nulls_distinct>` was added.
 
+.. versionchanged:: 6.0.4
+
+    More optimized unique constraint filters for exact lookups were added,
+    allowing the query planner to use indexes for ``string``, ``int``,
+    ``long``, ``double``, ``decimal``, ``bool``, and ``date`` fields. Unique
+    constraints that rely on ``$type`` filters aren't used by the query planner.
+
+
```
timgraham: Always one blank line in docs, never two.

docs/releases/6.0.x.rst
```diff
@@ -48,6 +48,9 @@ Bug fixes
   :class:`~django_mongodb_backend.fields.PolymorphicEmbeddedModelArrayField`
   that specifies ``embedded_models`` as a list of concrete models.
 
+- Ensure unique indexes for ``string``, ``int``, ``long``, ``double``,
+  ``decimal``, ``bool``, and ``date`` fields are being used by the query planner.
```
timgraham: - Follow the style of existing notes (past tense). 
- Avoid technical jargon (db types). Instead describe what model fields are affected.

docs/releases/6.0.x.rst
```diff
@@ -79,6 +82,10 @@ Backwards incompatible changes
   :setting:`INSTALLED_APPS` setting in order to generate migration operations
   for embedded models.
 
+- Users will need to drop and recreate unique indexes on ``string``, ``int``,
+  ``long``, ``double``, ``decimal``, ``bool``, and ``date`` fields in order to
+  benefit from the more optimized unique constraint filters.
```
timgraham: Existing style: "You" not "Users".
Unclear what "more optimized unique constraint filters" means. You described the query planner issue in the bug fix note. I would include everything here rather than putting it in separate sections.

docs/ref/models/constraints.rst
```diff
@@ -23,6 +23,14 @@ If you wish to only allow one document with a ``NULL`` value, use a
     Support for :attr:`UniqueConstraint.nulls_distinct
     <django.db.models.UniqueConstraint.nulls_distinct>` was added.
 
+.. versionchanged:: 6.0.4
+
+    More optimized unique constraint filters for exact lookups were added,
+    allowing the query planner to use indexes for ``string``, ``int``,
+    ``long``, ``double``, ``decimal``, ``bool``, and ``date`` fields. Unique
+    constraints that rely on ``$type`` filters aren't used by the query planner.
```
timgraham: These sorts of annotations should be "self-contained". That is, the documentation should be written in the normal flow, then the annotation should be added to explain what was added. The idea is that after a few release cycles, the versionchanged/added annotations can simply be removed without having to restructure the documentation.

(I'm going to leave a lot of documentation comments, but don't worry about getting everything perfect. I'll make some revisions afterward.)

tests/constraints_/test_unique_indexes.py
```diff
@@ -0,0 +1,214 @@
...
+class UniqueIndexTests(SimpleTestCase):
+    def test_single_field_unique_index_filter(self):
+        class Author(models.Model):
+            name = models.TextField(unique=True)
+
+            class Meta:
+                app_label = "constraints_"
+
+        field = Author._meta.get_field("name")
+        constraint = models.UniqueConstraint(fields=["name"], name="author_name_uniq")
```
timgraham: These tests are a bit confusing. Each field has `unique=True` as well as a `UniqueConstraint`. I'm not sure if these tests are really needed because they are testing an implementation detail (the index definition). The more important tests are that the query planner actually uses the definitions so I was expecting the lookup tests to be expanded. (If we keep these tests, probably they could be a single parameterized test as I suggest below for the lookup tests).

tests/lookup_/models.py
```diff
@@ -17,3 +17,21 @@ class Meta:
 
     def __str__(self):
         return str(self.num)
+
+
+class UniqueAuthor(models.Model):
+    name = models.TextField(unique=True)
```
timgraham: I would suggest a single "UniqueFields" model that has every affected field. Then you can transform `test_exact_lookup_uses_partial_unique_index` into a single parameterized test with [`subTest()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest) that covers all fields.

tests/lookup_/tests.py
```diff
@@ -170,3 +170,34 @@ def test_subquery_filter_constant(self):
...
+    def test_compound_exact_lookup_uses_partial_unique_index(self):
+        author = UniqueAuthor.objects.create(name="JK Rowling")
+        UniqueBook.objects.create(author=author, version=3, name="Harry Potter")
+        UniqueBook.objects.create(author=author, version=4, name="Harry Potter")
+
+        plan = json_util.loads(UniqueBook.objects.filter(version=3, name="Harry Potter").explain())[
+            "queryPlanner"
+        ]["winningPlan"]
+
+        self.assertTrue(self._plan_contains_ixscan(plan))
```
timgraham: It's unclear to me if the compound index needs to be tested. Do we expect the query planner to work differently for these cases? If so, which combinations of compound fields need to be tested? 

a nit aside: the author field isn't used, so avoid including it (if `UniqueBook` is even needed).

django_mongodb_backend/constraints.py
```diff
@@ -10,6 +13,44 @@
 from .indexes import EmbeddedFieldIndexMixin, _get_condition_mql, get_field
 
 
+def _get_partial_unique_filter(field, connection):
+    field = getattr(field, "field", field)
+    db_type = field.db_type(connection)
+
+    match db_type:
```
timgraham: I'm in the habit of omitting blank lines as much as possible. If I feel myself wanting to add a blank line to create logical separation, I find it's more usefully replaced by a comment.

tests/lookup_/tests.py
```diff
@@ -170,3 +170,34 @@ def test_subquery_filter_constant(self):
                 {"$sort": SON([("num", 1)])},
             ],
         )
+
+
+class IndexLookupTests(TestCase):
+    def _plan_contains_ixscan(self, plan):
+        if isinstance(plan, dict):
+            if "IXSCAN" in (plan.get("stage") or ""):
```
timgraham:
```suggestion
            if "IXSCAN" in plan.get("stage", ""):
```

tests/lookup_/tests.py
```diff
@@ -170,3 +170,34 @@ def test_subquery_filter_constant(self):
...
+                return True
+            return any(self._plan_contains_ixscan(value) for value in plan.values())
+        if isinstance(plan, list):
```
timgraham: Do we need this second check? Could `plan` be any other type than list/dict?

sophiayangDB: During recursion over `plan.values()`, it's possible to hit scalars like strings, numbers, booleans, etc.

django_mongodb_backend/constraints.py
sophiayangDB: Honestly not sure either, an artifact from Jib's JIRA comment.

tests/lookup_/tests.py
sophiayangDB: Noted, doesn't seem like it.

timgraham (approved):

PR #544, opened 2026-05-28
INTPYTHON-973 Add support for the Right database function

timgraham:


aclark4life:
Do we need a test for this?

timgraham:
`db_functions.text.test_right.RightTests` is removed from skipped tests.

aclark4life (approved):

PR #546, opened 2026-05-28
INTPYTHON-974 Add support for MD5/SHA256 database functions

timgraham:


aclark4life:
What's going on in the test, a `DTModel` instance gets an encrypted annotation right? Could a field of `DTModel` hold an encrypted value?

timgraham:
The functions compute a given value's hash. It's unrelated to queryable encryption and encrypted fields.

The tests are a bit bogus since the input of `$hash` "Must resolve to a valid UTF-8 string or BinData" (and here we have a datetime) but it's good enough to trigger the error being tested. SInce these tests are temporary until older versions of MongoDB are EOL, I didn't see any point in adding another model for these tests in the meantime. Django's test suite has tests with proper inputs.

aclark4life (approved):

PR #547, opened 2026-05-30
INTPYTHON-975 Add support for LPad, RPad, Repeat, Reverse database functions

timgraham:
I'm not sure how these will perform, but if there are any complaints, perhaps that will help make a case to the server team to implement specialized aggregation expressions.


aclark4life (approved):
Nice!

PR #549, opened 2026-06-02
INTPYTHON-792 Add support for window functions

timgraham:
Claude did the heavy lifting. I did a little refactoring, made some cosmetic edits, and removed unneeded code.

I added additional test coverage in Django's test suite: https://github.com/django/django/pull/21515

### DIFF
--- a/django_mongodb_backend/compiler.py
+++ b/django_mongodb_backend/compiler.py
@@ -1,4 +1,5 @@
 import itertools
+import json
 from collections import defaultdict
 
 from bson import SON, ObjectId, json_util
@@ -6,7 +7,7 @@ from django.core.exceptions import EmptyResultSet, FieldError, FullResultSet
 from django.db import IntegrityError, NotSupportedError
 from django.db.models import Count
 from django.db.models.aggregates import Aggregate, Variance
-from django.db.models.expressions import Case, Col, OrderBy, Ref, Value, When
+from django.db.models.expressions import Case, Col, OrderBy, Ref, RowRange, Value, When, Window
 from django.db.models.functions.comparison import Coalesce
 from django.db.models.functions.math import Power
 from django.db.models.lookups import IsNull
@@ -38,6 +39,8 @@ class SQLCompiler(compiler.SQLCompiler):
         self.subqueries = []
         # Search stage.
         self.search_pipeline = []
+        # Window function pipeline stages ($setWindowFields + post-$addFields).
+        self.window_pipeline = None
         # Does the aggregation have no GROUP BY fields and need wrapping?
         self.needs_wrap_aggregation = False
         # The MQL equivalent to a SQL HAVING clause.
@@ -212,6 +215,11 @@ class SQLCompiler(compiler.SQLCompiler):
         having_group_by = self.having.get_group_by_cols() if self.having else ()
         for expr in having_group_by:
             expressions.add(expr)
+        # Include partition/sort columns from window annotations so that
+        # $setWindowFields can partition the grouped data.
+        for expr in self.query.annotations.values():
+            if expr.contains_over_clause:
+                expressions |= set(expr.get_group_by_cols())
         return expressions
 
     def _get_group_id_expressions(self, order_by):
@@ -308,7 +316,8 @@ class SQLCompiler(compiler.SQLCompiler):
         extra_select, order_by, group_by = super().pre_sql_setup(with_col_aliases=with_col_aliases)
         search_replacements = self._prepare_search_query_for_aggregation_pipeline(order_by)
         group, group_replacements = self._prepare_annotations_for_aggregation_pipeline(order_by)
-        all_replacements = {**search_replacements, **group_replacements}
+        window_stages, window_replacements = self._prepare_window_annotations_for_pipeline()
+        all_replacements = {**search_replacements, **group_replacements, **window_replacements}
         self.search_pipeline = self._compound_searches_queries(search_replacements)
         # query.group_by is either:
         # - None: no GROUP BY
@@ -329,6 +338,7 @@ class SQLCompiler(compiler.SQLCompiler):
                 self.subqueries = []
                 pipeline.append({"$match": having})
             self.aggregation_pipeline = pipeline
+        self.window_pipeline = window_stages
         self.annotations = {
             target: expr.replace_expressions(all_replacements)
             for target, expr in self.query.annotation_select.items()
@@ -336,6 +346,10 @@ class SQLCompiler(compiler.SQLCompiler):
         self.order_by_objs = [expr.replace_expressions(all_replacements) for expr, _ in order_by]
         if (where := self.get_where()) and search_replacements:
             self.set_where(where.replace_expressions(search_replacements))
+        # Apply window/annotation replacements to the qualify clause so that
+        # Window expressions are replaced by their pre-computed field Refs.
+        if getattr(self, "qualify", None) and window_replacements:
+            self.qualify = self.qualify.replace_expressions(all_replacements)
         return extra_select, order_by, group_by
 
     def get_project_columns(self, columns):
@@ -493,6 +507,9 @@ class SQLCompiler(compiler.SQLCompiler):
         if extra_fields:
             query.extra_fields = self.get_project_fields(extra_fields, force_expression=True)
         query.subqueries = self.subqueries
+        query.window_pipeline = self.window_pipeline
+        if getattr(self, "qualify", None):
+            query.qualify_mql = self.qualify.as_mql(self, self.connection)
         return query
 
     @cached_property
@@ -800,6 +817,9 @@ class SQLCompiler(compiler.SQLCompiler):
     def _get_search_expressions(self, expr):
         return self._get_all_expressions_of_type(expr, SearchExpression)
 
+    def _get_window_expressions(self, expr):
+        return self._get_all_expressions_of_type(expr, Window)
+
     def _get_all_expressions_of_type(self, expr, target_type):
         stack = [expr]
         while stack:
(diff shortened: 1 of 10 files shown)


django_mongodb_backend/expressions/builtins.py
```diff
@@ -157,6 +158,11 @@ def query(self, compiler, connection, get_wrapping_pipeline=None, as_expr=False)
         else:
             if subquery.aggregation_pipeline is None:
                 subquery.aggregation_pipeline = []
+            # Window stages must precede the wrapping pipeline so the $group
+            # can reference the window function outputs.
+            if subquery.window_pipeline:
+                subquery.aggregation_pipeline.extend(subquery.window_pipeline)
```
NoahStapp, line 164: According to the comment, "window stages must precede the wrapping pipeline", but here we add the window stages to the end of the pipeline?

NoahStapp (changes requested):

django_mongodb_backend/expressions/builtins.py
timgraham: The wrapping pipeline is added in the line below the added code:
`subquery.aggregation_pipeline.extend(wrapping_result_pipeline)`.

django_mongodb_backend/compiler.py
```diff
@@ -809,6 +829,213 @@ def _get_all_expressions_of_type(self, expr, target_type):
...
+        """
+        Compile Window.order_by to a MongoDB sortBy document and any pre-sort
+        $addFields.
+        """
+        if order_by_list is None:
+            return {}, {}
+        exprs = order_by_list.get_source_expressions()
+        sort_doc = {}
+        pre_add_fields = {}
+        sort_idx = itertools.count(start=1)
```
NoahStapp: Two windows with different `order_by` expressions in the same partition will both have the same field names and cause a collision. For example: 

```
qs = Employee.objects.annotate(
            a=Window(
                Sum("salary"),
                partition_by="department",
                order_by=(F("age") + F("salary")).asc(),
            ),
            b=Window(
                Sum("salary"),
                partition_by="department",
                order_by=(F("salary") + F("hire_date")).asc(),
            ),
        ).order_by("name")
```

Both will use `__wsort1`.

django_mongodb_backend/functions/window.py
```diff
@@ -0,0 +1,170 @@
...
+    NthValue,
+    Ntile,
+    PercentRank,
+    Rank,
+    RowNumber,
+)
+
+
+def aggregate(self, compiler, connection, alias, idx, default_frame):  # noqa: ARG001
+    operator = self.function.lower()
```
NoahStapp: `Window(StdDev), `Window(Variance)`, and `Window(Count)` all produce invalid MQL:

```
Window(StdDev) -> {alias: {"$stddev_samp/stddev_pop": ... # not valid MQL operators
Window(StdDev) -> {alias: {"$var_samp/var_pop": ... # not valid MQL operators
Window(Count) -> {alias: {"$count": ... # must have a value of {}
```

django_mongodb_backend/functions/window.py
```diff
@@ -0,0 +1,170 @@
...
+    exprs = self.get_source_expressions()
+    mql = exprs[0].as_mql(compiler, connection, as_expr=True)
```
NoahStapp: `Window(Sum('salary', filter=Q(department='ENG')))` drops the filter here since only `expr[0]` is read.

django_mongodb_backend/functions/window.py
```diff
@@ -0,0 +1,170 @@
...
+                "if": {"$gte": [{"$size": f"${push_alias}"}, nth]},
+                "then": {"$arrayElemAt": [f"${push_alias}", nth - 1]},
+                "else": None,
+            }
+        }
+    }
+    return output, add_fields
+
+
+def ntile(self, compiler, connection, alias, idx, default_frame):  # noqa: ARG001
```
NoahStapp, line 107: SQL ntile puts extras in the first buckets, here we put extras in the last buckets. Is that difference significant and if so do should we document or fix it?

django_mongodb_backend/functions/window.py
```diff
@@ -0,0 +1,170 @@
...
+        mql["default"] = exprs[2].as_mql(compiler, connection, as_expr=True)
+    return {alias: {"$shift": mql}}, {}
+
+
+def nth_value(self, compiler, connection, alias, idx, default_frame):  # noqa: ARG001
+    expr, nth_expr = self.get_source_expressions()
+    mql = expr.as_mql(compiler, connection, as_expr=True)
+    nth = nth_expr.value
+    push_alias = f"__wtemp{next(idx)}"
+    output = {push_alias: {"$push": mql, "window": {"documents": ["unbounded", "current"]}}}
```
NoahStapp: Hardcoded `window` means any user-passed `Window.frame` is ignored here.

django_mongodb_backend/expressions/builtins.py
NoahStapp: A subquery with `__in` will cause a misordering here:

```
  top_per_dept = (
      Employee.objects.annotate(r=Window(Rank(), order_by=F("salary").desc()))
      .filter(r=1)
      .values("name")
  )

result = list(
    Employee.objects.filter(name__in=Subquery(top_per_dept))
    .values_list("name", flat=True)
)

# causes this DatabaseError
pymongo.errors.OperationFailure: Executor error during aggregate command on namespace: test_djangotests_windowbugs.window_bugs__employee :: caused by :: $in requires an array as a second argument, found: missing, full error: {'ok': 0.0, 'errmsg': 'Executor error during aggregate command on namespace: test_djangotests_windowbugs.window_bugs__employee :: caused by :: $in requires an array as a second argument, found: missing', 'code': 40081, 'codeName': 'Location40081'}
```

django_mongodb_backend/compiler.py
```diff
@@ -809,6 +829,213 @@ def _get_all_expressions_of_type(self, expr, target_type):
...
+            else:
+                # item an an expression without an OrderBy wrapper.
+                expr = item
+                direction = ASCENDING
+            if isinstance(expr, (Col, Ref)):
+                field_name = expr.as_mql(self, self.connection, as_expr=False)
+            else:
+                field_name = f"__wsort{next(sort_idx)}"
+                pre_add_fields[field_name] = expr.as_mql(self, self.connection, as_expr=True)
+            sort_doc[field_name] = direction
```
NoahStapp: `nulls_first` and `nulls_last` aren't respected here:

```
qs = (
                Employee.objects.annotate(
                    r=Window(Rank(), order_by=F("hire_date").asc(nulls_last=True))
                )
                .filter(name="nullhire")
                .values_list("r", flat=True)
            )
# nulls are first
```

django_mongodb_backend/compiler.py
```diff
@@ -809,6 +829,213 @@ def _get_all_expressions_of_type(self, expr, target_type):
...
+                output_entries, post_entries = self._window_function_output(
+                    alias, window, idx, use_full_partition
+                )
+                set_window_fields_output.update(output_entries)
+                post_add_fields.update(post_entries)
+            # $documentNumber requires exactly one sortBy field; add _id as
+            # the default.
+            sort_doc = group_info["sort"]
+            if not sort_doc and any(
+                isinstance(v, dict) and "$documentNumber" in v
```
NoahStapp: This fallback doesn't catch other operators that also require a sortBy field:

```
list(Employee.objects.annotate(r=Window(Rank()))))

# Produces this error
pymongo.errors.OperationFailure: $rank must be specified with a top level sortBy expression with exactly one element, full error: {'ok': 0.0, 'errmsg': '$rank must be specified with a top level sortBy expression with exactly one element', 'code': 5371602, 'codeName': 'Location5371602'}
```

NoahStapp (changes requested):

django_mongodb_backend/compiler.py
timgraham: Added `test_multiple_windows_different_complex_order_by` in https://github.com/django/django/pull/21515.

django_mongodb_backend/functions/window.py
timgraham: Added `test_window_count`, `test_window_stddev`, `test_window_variance`.

django_mongodb_backend/functions/window.py
timgraham: Added `test_aggregate_filter`, `test_stddev_filter`, `test_variance_filter`.

django_mongodb_backend/functions/window.py
timgraham: Fixed. Added `test_ntile_uneven_distribution`.

django_mongodb_backend/functions/window.py
timgraham: Added `test_nthvalue_full_partition_frame`.

django_mongodb_backend/compiler.py
timgraham: Added `test_window_order_by_nulls_last` and `test_window_order_by_nulls_first`.

django_mongodb_backend/compiler.py
timgraham: Added `test_rank_no_order_by`, `test_dense_rank_no_order_by`.

django_mongodb_backend/expressions/builtins.py
timgraham: Added `test_window_qualify_filter_in_subquery`.

django_mongodb_backend/compiler.py
```diff
@@ -809,6 +829,248 @@ def _get_all_expressions_of_type(self, expr, target_type):
...
+        # $setWindowFields.
+        groups = {}
+        pre_add_fields = {}
+        sort_fields = {}  # shared across all windows, expr_mql_key: field_name
+        for alias, window in window_list:
+            partition_mql = self._compile_window_partition_by(window.partition_by)
+            sort_doc, pre_add_fields_ = self._compile_window_sort_by(window.order_by, sort_fields)
+            pre_add_fields.update(pre_add_fields_)
+            partition_key = json.dumps(partition_mql)
+            sort_key = json.dumps(sort_doc)
```
NoahStapp, line 1006: These will crash if `partition_mql` or `sort_doc` contain a BSON type like `Decimal128`:

```
v = Value(decimal.Decimal("1.5"))
qs = list(Employee.objects.annotate(
    r=Window(Rank(), partition_by=v, order_by=F("salary").asc()),
))
--------
    raise TypeError(f'Object of type {o.__class__.__name__} '
                    f'is not JSON serializable')
TypeError: Object of type Decimal128 is not JSON serializable
```
We should use `bson.json_util.dumps` instead.

django_mongodb_backend/compiler.py
```diff
@@ -809,6 +829,248 @@ def _get_all_expressions_of_type(self, expr, target_type):
...
+            else:
+                # item an an expression without an OrderBy wrapper.
+                expr = item
+                direction = ASCENDING
+            if isinstance(expr, (Col, Ref)):
+                field_name = expr.as_mql(self, self.connection, as_expr=False)
+                val_mql = expr.as_mql(self, self.connection, as_expr=True)
+            else:
+                val_mql = expr.as_mql(self, self.connection, as_expr=True)
+                expr_key = json.dumps(val_mql, sort_keys=True)
```
NoahStapp, line 868: Same here.

django_mongodb_backend/compiler.py
```diff
@@ -809,6 +829,248 @@ def _get_all_expressions_of_type(self, expr, target_type):
...
+
+        When sort == partition, all rows in the partition are tied in sort
+        order, so the correct SQL RANGE window is the full partition (unbounded
+        both ways).
+        """
+        if not partition_mql or not sort_doc:
+            return False
+
+        def dump(expr):
+            return json.dumps(expr, sort_keys=True)
```
NoahStapp, line 908: Same here.

NoahStapp (changes requested):

django_mongodb_backend/compiler.py
timgraham: Is it a reasonable use case? Claude says, "`PARTITION BY <constant>` puts every row into the same partition, so it's equivalent to no partition_by at all." Similarly, I can't think of a use case for ordering by a fixed Decimal.

NoahStapp: I would prefer to err on the safe side and not ship code that has a known path to get an internal `TypeError`, even if a user has to be doing something unreasonable to get there.

timgraham: Even with `json_util.dumps()` they could pass some other garbage like `object()` which fails in the same way. I have no problem with using `json_util.dumps()`, but I don't think I can make a reasonable case to include that regression test in Django's test suite. If you feel this is important enough to have additional window tests in this repo, I will do that.

NoahStapp: Fair enough. We can easily fix this in the future if users actually do it.

timgraham: I believe the same reasoning applies to all three `json.dumps()`. Do you have any more concerns or want to give this another review? Thanks.

NoahStapp (approved):

PR #551, opened 2026-06-05
INTPYTHON-978 Make embedded models unmanaged

timgraham:


aclark4life:
Before this did we have a noop somewhere to skip creating collections for embedded models ?

timgraham:
Two things:
- [@ignore_embedded_models](https://github.com/mongodb/django-mongodb-backend/blob/742c97b3a9d963e22070bc8a80137434096aa633/django_mongodb_backend/schema.py#L16-L31) decorator on `SchemaEditor` methods.
  - It remains to protect models in existing migrations (but could be removed after a deprecation that alerts users to update their old migrations).
- [MongoRouter.allow_migrate()](https://github.com/mongodb/django-mongodb-backend/blob/main/django_mongodb_backend/routers.py#L6-L18) method.
  - It remains to prevent embedded models from appearing in fixtures generated by `dumpdata`. Unmanaged models are dumped [by design](https://code.djangoproject.com/ticket/13816).

aclark4life (approved):
LGTM

PR #552, opened 2026-06-05
INTPYTHON-983 Add support for pattern lookups on UUIDField

timgraham:


aclark4life:
What is the context here? `UUIDField` is a Django field that we've support from the beginning correct? Is this due to a new server feature or are we just getting around to it? Do we have pattern matching on any other fields?

timgraham:
The implementation uses similar logic as [`UUIDTextMixin.process_rhs()`](https://github.com/django/django/blob/a7a57a21a8cf42ecc767c204f7939ad4419ae25f/django/db/models/lookups.py#L778-L794). I didn't come up with a solution previously. 

Most [other pattern matching](https://github.com/mongodb/django-mongodb-backend/blob/c950b21c03dbc5a309d9054b1390a07597260c9a/django_mongodb_backend/base.py#L153-L161) works.

aclark4life (approved):

PR #570, opened 2026-06-15
INTPYTHON-643 Prevent repeated errors on nested embedded forms

timgraham:
INTPYTHON-643

Claude's analysis:

**Root cause**

`forms.EmbeddedModelField` subclasses `forms.MultiValueField`. That single choice causes every symptom in the ticket:

1. Messages collapse onto the parent. `MultiValueField` is designed to represent one logical value split across several widgets (think split date/time). By contract, all subfield errors are aggregated into the parent field's errorlist — so `form.errors["publisher"]` gets `["Ensure this value has at most 2 characters (it has 8)."]` and it renders in a single `id_publisher_error` `<ul>` at the fieldset level, with no indication that state was the culprit.
2. Wrong/no field highlighting. Rendering is hijacked by `EmbeddedModelBoundField.__str__()`, which builds a fresh, unbound model form: `self.field.model_form_cls(instance=self.value(), ...)`. Because it's unbound (`instance=`, never `data=`), the nested form has no errors of its own, so Django can't add aria-invalid/error classes to the offending inputs. The fields that are actually wrong are never marked.
  3. Generic, useless messages. `"Enter all required values."` / `"Enter a list of values."` are `MultiValueField`'s incomplete/invalid messages — they don't name the missing subfield.

The same bug exists, in worse form, in `EmbeddedModelArrayField`: `clean()` raises `ValidationError(formset.errors + formset.non_form_errors())` (aggregating onto the parent), and `EmbeddedModelArrayBoundField.__str__()` renders a bound formset that shows per-field errors too → messages are literally rendered twice.

**Why this is the wrong abstraction**

`MultiValueField`/`MultiWidget` has no concept of per-subwidget errors — there is no template hook to attach an error to widget N. The structure that does have native per-field error association is a bound form (or formset): each `BoundField` renders its own errors next to its input and marks the input invalid. Django already does this perfectly; we just aren't letting it.

**Recommended solution**

Stop using `MultiValueField`. Make `EmbeddedModelField` a plain `forms.Field` that owns a real, bound nested `ModelForm` and delegates both validation and rendering to it — mirroring (and then unifying with) the `EmbeddedModelArrayField`/formset design, which is already the right shape.

The enabling hook is `Field._clean_bound_field(self, bf)`. Unlike `clean(value)`, it receives the `BoundField` — so the same nested form built for rendering can be the one we validate. That gives a single source of truth and lets us delete the prefix_override HACK.

---

This improves the situation and solves INTPYTHON-993 as a side-effect (commit two has regression tests), but it doesn't solve all layout and styling issues. I created INTPYTHON-995 for follow up.

### DIFF
--- a/django_mongodb_backend/forms/fields/embedded_model.py
+++ b/django_mongodb_backend/forms/fields/embedded_model.py
@@ -1,81 +1,114 @@
+from itertools import chain
+
 from django import forms
+from django.core.exceptions import ValidationError
 from django.forms.models import modelform_factory
 from django.utils.safestring import mark_safe
-from django.utils.translation import gettext_lazy as _
 
 
-class EmbeddedModelField(forms.MultiValueField):
-    default_error_messages = {
-        "invalid": _("Enter a list of values."),
-        "incomplete": _("Enter all required values."),
-    }
+class EmbeddedModelField(forms.Field):
+    """
+    Form field for an embedded model, backed by a ModelForm.
+
+    A single nested form instance (built in EmbeddedModelBoundField) is the
+    source of truth for both validation and rendering, so each validation error
+    renders inline next to the subfield that produced it instead of being
+    aggregated onto this parent field. The nested form's messages are still
+    aggregated into the parent form's errors dict (so form.errors[name] stays
+    meaningful), but EmbeddedModelBoundField.errors suppresses them at render
+    time to avoid showing them twice.
+    """
 
     def __init__(self, model, prefix, *args, **kwargs):
-        form_kwargs = {}
-        # To avoid collisions with other fields on the form, each subfield must
-        # be prefixed with the name of the field.
-        form_kwargs["prefix"] = prefix
-        self.form_kwargs = form_kwargs
+        self.model = model
+        # To avoid collisions with other fields on the form, each subfield is
+        # prefixed with the name of this field.
+        self.prefix = prefix
         self.model_form_cls = modelform_factory(model, fields="__all__")
-        self.model_form = self.model_form_cls(**form_kwargs)
-        self.field_names = list(self.model_form.fields.keys())
-        fields = self.model_form.fields.values()
-        widgets = [field.widget for field in fields]
-        widget = EmbeddedModelWidget(self.field_names, widgets)
-        super().__init__(*args, fields=fields, widget=widget, require_all_fields=False, **kwargs)
-
-    def compress(self, data_dict):
-        if not data_dict:
-            return None
-        values = dict(zip(self.field_names, data_dict, strict=False))
-        return self.model_form._meta.model(**values)
+        kwargs["widget"] = EmbeddedModelWidget(prefix)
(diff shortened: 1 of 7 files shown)


blink1073:
As a general question, is there a standard practice to use something like playwright with screenshots and error message handling as integration tests for Django UI?

timgraham:
- Django's [LiveServerTestCase](https://docs.djangoproject.com/en/6.0/topics/testing/tools/#django.test.LiveServerTestCase) can be used with Selenium, [Playright](https://github.com/mxschmitt/python-django-playwright), etc.
- [AdminSeleniumTestCase](https://github.com/django/django/blob/61e078ecab9a9e8de9d283adf9fa554b6a145ce8/django/contrib/admin/tests.py#L25) is a private API used by the Django test suite.
-  There's an ongoing 2026 Summer of Code project to [switch `AdminSeleniumTestCase` from Selenium to Playwright](https://summerofcode.withgoogle.com/programs/2026/projects/hVM2skuP).

blink1073:
Thanks, I created [INTPYTHON-1000 - [Django] Look into integration testing of the UI](https://jira.mongodb.org/browse/INTPYTHON-1000)

django_mongodb_backend/forms/fields/embedded_model.py
```diff
@@ -1,81 +1,114 @@
...
-        # When rendering a form with errors, nested EmbeddedModelField data
-        # won't be compressed if MultiValueField.clean() raises ValidationError
-        # error before compress() is called. The data must be compressed here
-        # so that EmbeddedModelBoundField.value() returns a model instance
-        # (rather than a list) for initializing the form in
-        # EmbeddedModelBoundField.__str__().
-        return self.compress(value) if isinstance(value, list) else value
+        return EmbeddedModelBoundField(form, self, field_name)
+
+    def _clean_bound_field(self, bf):
```
blink1073, line 34: Using a private API feels off, is this something we should try to refactor upstream?

timgraham: `_clean_bound_field()` was introduced in https://github.com/django/django/commit/d9b91e38361696014bdc98434d6d018eae809519. I don't see an obvious way to argue for a public API there (already called from a private API). Is it likely to change? Hard to say, but surely some way to accomplish the desired result will remain. As I explained in yesterday's meeting, periodically checking compatibility with Django's main branch is helpful so that if a breaking change is made, this project's maintainers can have an early dialog with Django's maintainers if needed.

The entire embedded model initiative stretches Django's APIs in new and explored ways. Absent a grand rearchitecture that abandons Django's usual APIs, I think it's likely to remain a bit of a kludge.

blink1073: Fair enough, given that we only ever target a single minor version, I think we have to accept the usage of some private APIs.

blink1073 (approved):
LGTM

PR #572, opened 2026-06-19
Use EmbeddedModelField.contribute_to_class() to resolve embedded_model

timgraham:
This follows the usual Django pattern. The _set_model() approach (adapted from djangotoolbox.fields.EmbeddedModelField) did the same thing but in an usual way.


blink1073 (approved):
LGTM

PR #574, opened 2026-06-22
INTPYTHON-992 Make Sum return None instead of 0 when no rows match

timgraham:
Example fixed queries:

1. Publisher "Jonno's House of Books" has no books, so currently returns 0 instead of None.
```python
publishers = Publisher.objects.annotate(Sum("book__price")).order_by("name")
self.assertQuerySetEqual(
    publishers,
    [
        ("Apress", Decimal("59.69")),
        ("Jonno's House of Books", None),
        ("Morgan Kaufmann", Decimal("75.00")),
        ("Prentice Hall", Decimal("112.49")),
        ("Sams", Decimal("23.09")),
    ],
    lambda p: (p.name, p.book__price__sum),
)
```

2. Currently "Jonno's House of Books" does not appear in the results since its `rating_sum` is 0 instead of null.
```python
qs = (
    Publisher.objects.annotate(
        rating_sum=Sum("book__rating"), book_count=Count("book")
    )
    .filter(Q(rating_sum__gt=5.5) | Q(rating_sum__isnull=True))
    .order_by("pk")
)
self.assertQuerySetEqual(
    qs,
    [
        "Apress",
        "Prentice Hall",
        "Jonno's House of Books",
    ],
    attrgetter("name"),
)
````

### DIFF
--- a/django_mongodb_backend/aggregates.py
+++ b/django_mongodb_backend/aggregates.py
@@ -4,6 +4,7 @@ from django.db.models.aggregates import (
     Count,
     StdDev,
     StringAgg,
+    Sum,
     Variance,
 )
 from django.db.models.expressions import Case, Value, When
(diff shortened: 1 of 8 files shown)


aclark4life (approved):
LGTM

PR #576, opened 2026-06-23
INTPYTHON-994 Make QuerySet.exclude() omit documents with absent JSONField keys

timgraham:
`exclude(value__key=X)` and negated Q objects on `JSONField` key lookups wrongly returned documents that don't contain the key. SQL databases treat the comparison on a missing key as `NULL` (three-valued logic), so `NOT(...)` excludes those rows; MongoDB's `$nor`/`$not` instead treat a missing key as "not equal" and included them.

Give `KeyTransformExact` a custom negation (key exists AND value != RHS), and push negation down to the leaves via De Morgan's laws in `where_node()` when a subtree contains such a lookup, so each lookup can supply its own SQL-correct negation while other lookups keep using `$nor`/`$not`.

### DIFF
--- a/django_mongodb_backend/fields/json.py
+++ b/django_mongodb_backend/fields/json.py
@@ -15,6 +15,7 @@ from django.db.models.fields.json import (
     KeyTransformIn,
     KeyTransformIsNull,
     KeyTransformNumericLookupMixin,
+    KeyTransformTextLookupMixin,
 )
 
 from django_mongodb_backend.lookups import builtin_lookup_expr, builtin_lookup_path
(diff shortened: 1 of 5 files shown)


django_mongodb_backend/query.py
```diff
@@ -355,7 +355,45 @@ def _get_reroot_replacements(expression):
...
+            return negate(compiler, connection, as_expr=as_expr)
+        mql = node.as_mql(compiler, connection, as_expr=as_expr)
+        return {"$not": [mql]} if as_expr else {"$nor": [mql]}
+    if node.negated:
+        # NOT(NOT(group)) == group; compile the group without negation.
+        node = node.__class__(node.children, node.connector)
+        return node.as_mql(compiler, connection, as_expr=as_expr)
+    operator = "$or" if node.connector == AND else "$and"
+    parts = [_negate(child, compiler, connection, as_expr) for child in node.children]
+    return parts[0] if len(parts) == 1 else {operator: parts}
```
NoahStapp, line 413: What if `parts` is empty? Should that give us no rows or all rows?

django_mongodb_backend/fields/json.py
```diff
@@ -228,6 +250,7 @@ def register_json_field():
     KeyTransform.can_use_path = key_transform_is_simple_column
     KeyTransform.is_simple_column = key_transform_is_simple_column
     KeyTransformExact.as_mql_path = key_transform_exact_path
+    KeyTransformExact.as_mql_negated = key_transform_exact_negated
```
NoahStapp: This only covers `KeyTransformExact`: do other JSON lookups (gt, lt, gte, lte, in, regex) not have the same issue?

NoahStapp (dismissed):

django_mongodb_backend/query.py
timgraham: I believe it won't happen:

1. Entry from `where_node()` (line 396): guarded by `_has_custom_negation(self)`, which is `any(... for child in node.children)` — that returns False for a childless node, so we never enter `_negate()` with an empty top node. The entry node always has ≥1 child.
  2. Recursive descent (line 388 iterating into a non-negated child `WhereNode`): Django prunes empty `Q(`)/`WhereNode` objects during query construction. For example, `exclude(Q(value__c__gt=10) & Q())` doesn't produce an empty child node; the empty `Q()` is dropped and the tree is just the lookup plus the `IsNull` term Django adds for negated key lookups.

django_mongodb_backend/fields/json.py
timgraham: They do. Fixed and added `test_lookup_exclude_other_lookups`. https://github.com/django/django/pull/21567

django_mongodb_backend/query.py
NoahStapp: So something like `Document.objects.exclude(value__k="x", num__in=[])` should return everything? For me it returns nothing.

timgraham: You are correct. I fixed this and added a second commit to https://github.com/django/django/pull/21567 to cover the new code: "Added JSONKey key exclude() tests for Empty/FullResultSet cases."

NoahStapp: I still see no documents running `Document.objects.exclude(value__k__in=[])`.

django_mongodb_backend/query.py
```diff
@@ -361,7 +361,63 @@ def _get_reroot_replacements(expression):
...
+        except EmptyResultSet as exc:
+            raise FullResultSet from exc
+        except FullResultSet as exc:
+            raise EmptyResultSet from exc
+        return {"$not": [mql]} if as_expr else {"$nor": [mql]}
+    if node.negated:
+        # NOT(NOT(group)) == group; compile the group without negation.
+        node = node.__class__(node.children, node.connector)
+        return node.as_mql(compiler, connection, as_expr=as_expr)
+    operator = "$or" if node.connector == AND else "$and"
```
NoahStapp, line 398: Is it possible for a nested XOR to hit this and incorrectly negate?

NoahStapp (changes requested):

timgraham:
I'm out of time with MongoDB, so I'l have to hand this off.

PR #579, opened 2026-06-23
Remove statement that database caching may be added in a future release

timgraham:
INTPYTHON-451 was closed as "won't do" because Django's database caching uses pickle to serialize cached data and this could escalate a cache compromise into a remote-code execution. Mitigations such as signing cached data could be implemented but this has performance implications.

### DIFF
--- a/docs/topics/known-issues.rst
+++ b/docs/topics/known-issues.rst
@@ -97,6 +97,4 @@ Due to the lack of ability to introspect MongoDB collection schema,
 Caching
 =======
 
-:ref:`Database caching <database-caching>` is not supported since Django's built-in
-database cache backend requires SQL. A custom cache backend for MongoDB may be provided
-in the future.
+:ref:`Database caching <database-caching>` is not supported.


docs/topics/known-issues.rst
```diff
@@ -101,6 +101,5 @@ Due to the lack of ability to introspect MongoDB collection schema,
 Caching
 =======
 
-:ref:`Database caching <database-caching>` is not supported since Django's built-in
-database cache backend requires SQL. A custom cache backend for MongoDB may be provided
-in the future.
+:ref:`Database caching <database-caching>` is not supported because Django's
+built-in database cache backend requires SQL.
```
aclark4life: Should we say any more about why no custom cache backend for MongoDB?

timgraham: It boils down to "The MongoDB Security Team disagrees with the Django team's design decision that a cache backend that allows a database compromise to escalate into an RCE is acceptable." What do you think?

aclark4life: > It boils down to "The MongoDB Security Team disagrees with the Django team's design decision that a cache backend that allows a database compromise to escalate into an RCE is acceptable." What do you think?

```
:ref:`Database caching <database-caching>` is not supported
due to the inherent insecurity of available caching methods.
```

Also, I still think put this in extensions may be worth pursuing …

timgraham: I think a vague statement like that isn't helpful. I would not characterize it as inherently insecure (unless you consider your database insecure). A compromised database leads to a host of problems. Consider [the warning](https://docs.djangoproject.com/en/6.0/topics/cache/#:~:text=An%20attacker%20who%20gains%20access%20to%20the%20cache%20file%20can%20not%20only%20falsify%20HTML%20content%2C%20which%20your%20site%20will%20trust%2C%20but%20also%20remotely%20execute%20arbitrary%20code%2C%20as%20the%20data%20is%20serialized%20using%20pickle%2E) in the Django docs: 

> An attacker who gains access to the cache file can not only falsify HTML content, which your site will trust, but also remotely execute arbitrary code, as the data is serialized using [pickle](https://docs.python.org/3/library/pickle.html#module-pickle).

As I have said before, my opinion is that putting a cache backend in the the extensions library only makes maintenance more difficult.

aclark4life (approved):

PR #580, opened 2026-06-24
INTPYTHON-917 - Add support for flat indexing method in vector search index definitions

NoahStapp:
[INTPYTHON-917](https://jira.mongodb.org/browse/INTPYTHON-917)

## Changes in this PR

Adds support for `indexingMethod` to `VectorSearchIndexModel` for "flat" index support with MongoDB Vector Search.

## Test Plan

Added new unit and integration tests consistent with how the existed `similarities` option is tested.

### Checklist for Author

- [X] Did you update the changelog (if necessary)?
- [X] Is the intention of the code captured in relevant tests?
- ~[ ] If there are new TODOs, has a related JIRA ticket been created?~

### Checklist for Reviewer

### DIFF
--- a/django_mongodb_backend/indexes.py
+++ b/django_mongodb_backend/indexes.py
@@ -308,8 +308,9 @@ class VectorSearchIndex(SearchIndex):
     suffix = "vsi"
     VALID_FIELD_TYPES = frozenset(("boolean", "date", "number", "objectId", "string", "uuid"))
     VALID_SIMILARITIES = frozenset(("cosine", "dotProduct", "euclidean"))
+    VALID_INDEXING_METHODS = frozenset(("hnsw", "flat"))
 
-    def __init__(self, *, fields=(), name=None, similarities):
+    def __init__(self, *, fields=(), name=None, similarities, indexing_methods=None):
         super().__init__(fields=fields, name=name)
         self.similarities = similarities
         self._multiple_similarities = isinstance(similarities, (tuple, list))
@@ -319,6 +320,17 @@ class VectorSearchIndex(SearchIndex):
                     f"'{func}' isn't a valid similarity function "
                     f"({', '.join(sorted(self.VALID_SIMILARITIES))})."
                 )
+        self.indexing_methods = indexing_methods
+        self._multiple_indexing_methods = isinstance(indexing_methods, (tuple, list))
+        if indexing_methods is not None:
+            for method in (
+                indexing_methods if self._multiple_indexing_methods else (indexing_methods,)
+            ):
+                if method not in self.VALID_INDEXING_METHODS:
+                    raise ValueError(
+                        f"'{method}' isn't a valid indexing method "
+                        f"({', '.join(sorted(self.VALID_INDEXING_METHODS))})."
+                    )
         seen_fields = set()
         for field_name, _ in self.fields_orders:
             if field_name in seen_fields:
@@ -376,6 +388,17 @@ class VectorSearchIndex(SearchIndex):
                     id="mongodb.indexes.search.E005",
                 )
             )
+        if self._multiple_indexing_methods and num_arrayfields != len(self.indexing_methods):
+            errors.append(
+                Error(
+                    f"VectorSearchIndex requires the same number of indexing methods "
+                    f"and vector fields; {model._meta.object_name} has "
+                    f"{num_arrayfields} ArrayField(s) but indexing_methods "
+                    f"has {len(self.indexing_methods)} element(s).",
+                    obj=model,
+                    id="mongodb.indexes.search.E007",
+                )
+            )
         if num_arrayfields == 0:
             errors.append(
                 Error(
@@ -391,6 +414,8 @@ class VectorSearchIndex(SearchIndex):
     def deconstruct(self):
         path, args, kwargs = super().deconstruct()
         kwargs["similarities"] = self.similarities
+        if self.indexing_methods is not None:
+            kwargs["indexing_methods"] = self.indexing_methods
         return path, args, kwargs
 
     def get_pymongo_index_model(self, model, schema_editor, field=None, column_prefix=""):
@@ -400,6 +425,11 @@ class VectorSearchIndex(SearchIndex):
             itertools.cycle([self.similarities])
             if not self._multiple_similarities
             else iter(self.similarities)
+        )
+        indexing_methods = (
+            itertools.cycle([self.indexing_methods])
+            if not self._multiple_indexing_methods
+            else iter(self.indexing_methods)
         )
         fields = []
         for field_name, _ in self.fields_orders:
@@ -414,6 +444,9 @@ class VectorSearchIndex(SearchIndex):
                         "similarity": next(similarities),
                     }
                 )
+                method = next(indexing_methods)
+                if method is not None:
+                    mappings["indexingMethod"] = method
             else:
                 mappings["type"] = "filter"
             fields.append(mappings)
--- a/django_mongodb_backend/indexes.py
+++ b/django_mongodb_backend/indexes.py
@@ -388,6 +388,16 @@ class VectorSearchIndex(SearchIndex):
                     id="mongodb.indexes.search.E005",
                 )
             )
+        if num_arrayfields == 0:
+            errors.append(
+                Error(
+                    "VectorSearchIndex requires at least one ArrayField to store vector data.",
+                    obj=model,
+                    id="mongodb.indexes.search.E006",
+                    hint="If you want to perform search operations without vectors, "
+                    "use SearchIndex instead.",
+                )
+            )
         if self._multiple_indexing_methods and num_arrayfields != len(self.indexing_methods):
             errors.append(
                 Error(
@@ -397,16 +407,6 @@ class VectorSearchIndex(SearchIndex):
                     f"has {len(self.indexing_methods)} element(s).",
                     obj=model,
                     id="mongodb.indexes.search.E007",
-                )
-            )
-        if num_arrayfields == 0:
-            errors.append(
-                Error(
-                    "VectorSearchIndex requires at least one ArrayField to store vector data.",
-                    obj=model,
-                    id="mongodb.indexes.search.E006",
-                    hint="If you want to perform search operations without vectors, "
-                    "use SearchIndex instead.",
                 )
             )
         return errors
@@ -426,11 +426,14 @@ class VectorSearchIndex(SearchIndex):
             if not self._multiple_similarities
             else iter(self.similarities)
         )
-        indexing_methods = (
-            itertools.cycle([self.indexing_methods])
-            if not self._multiple_indexing_methods
-            else iter(self.indexing_methods)
-        )
+        if self.indexing_methods is None:
+            indexing_methods = None
+        else:
+            indexing_methods = (
+                iter(self.indexing_methods)
+                if self._multiple_indexing_methods
+                else itertools.cycle([self.indexing_methods])
+            )
         fields = []
         for field_name, _ in self.fields_orders:
             field_ = model._meta.get_field(field_name)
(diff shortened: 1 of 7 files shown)


django_mongodb_backend/indexes.py
```diff
@@ -414,6 +447,8 @@ def get_pymongo_index_model(self, model, schema_editor, field=None, column_prefi
                         "similarity": next(similarities),
                     }
                 )
+                if indexing_methods is not None:
+                    mappings["indexingMethod"] = next(indexing_methods)
```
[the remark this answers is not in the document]
NoahStapp: This is a valid point: do we want to depend on the server's default behavior and assume it won't change in the future? Attaching `hnsw` here for explicitness protects against a default change.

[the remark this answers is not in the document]
blink1073: The MongoDB [docs](https://www.mongodb.com/docs/vector-search/index/vector-search-type/?deployment-type=None&embedding=auto&interface=None&language=None#std-label-mdb-vs-indexing-method-auto) call out `hnsw` as the default, so changing the default would be a breaking change.  I think it is safe to omit.

django_mongodb_backend/indexes.py
```diff
@@ -319,6 +320,17 @@ def __init__(self, *, fields=(), name=None, similarities):
...
+        self._multiple_indexing_methods = isinstance(indexing_methods, (tuple, list))
+        if indexing_methods is not None:
+            for method in (
+                indexing_methods if self._multiple_indexing_methods else (indexing_methods,)
+            ):
+                if method not in self.VALID_INDEXING_METHODS:
+                    raise ValueError(
+                        f"'{method}' isn't a valid indexing method "
+                        f"({', '.join(sorted(self.VALID_INDEXING_METHODS))})."
+                    )
```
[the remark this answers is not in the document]
blink1073: This is true, we don't have type annotation so we may want to validate the input

docs/ref/checks.rst
```diff
@@ -37,6 +37,10 @@ Indexes
   :class:`.VectorSearchIndex` requires at least one :class:`.ArrayField` to
   store vector data. If you want to perform search operations without vectors,
   use :class:`.SearchIndex` instead.
+* **mongodb.indexes.search.E007**:
+  :class:`.VectorSearchIndex` requires the same number of indexing methods and
+  vector fields; ``<model>`` has  ``<#>`` ``ArrayField``\(s) but indexing_methods
+  has ``<#>``  element(s).
```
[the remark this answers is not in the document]
blink1073: I agree with this one

django_mongodb_backend/indexes.py
```diff
@@ -319,6 +320,17 @@ def __init__(self, *, fields=(), name=None, similarities):
                     f"'{func}' isn't a valid similarity function "
                     f"({', '.join(sorted(self.VALID_SIMILARITIES))})."
                 )
+        self.indexing_methods = indexing_methods
+        self._multiple_indexing_methods = isinstance(indexing_methods, (tuple, list))
+        if indexing_methods is not None:
+            for method in (
```
blink1073: I think the ternary assignment should be before the `for` block

django_mongodb_backend/indexes.py
```diff
@@ -401,6 +426,14 @@ def get_pymongo_index_model(self, model, schema_editor, field=None, column_prefi
             if not self._multiple_similarities
             else iter(self.similarities)
         )
+        if self.indexing_methods is None:
+            indexing_methods = None
+        else:
+            indexing_methods = (
```
blink1073, line 433: I think self.indexing_methods should always be a `tuple` or `None`

docs/conf.py
```diff
@@ -59,6 +59,7 @@
 linkcheck_ignore = [
```
blink1073: This should be changed to [linkcheck_anchors_ignore](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_anchors_ignore) instead of not testing the link

blink1073 (changes requested):

django_mongodb_backend/indexes.py
NoahStapp: You mean instead of a tuple, list, str, or None?

django_mongodb_backend/indexes.py
NoahStapp: Something like this?

```
if indexing_methods is not None:
    indexing_methods = indexing_methods if self._multiple_indexing_methods else (indexing_methods,)
    for method in indexing_methods:
        ...
```

Is this a readability concern, or something else?

django_mongodb_backend/indexes.py
blink1073: It is also currently a string in some cases, which is extra confusing, because you can iterate on a string and get characters

django_mongodb_backend/indexes.py
blink1073: readability, I had to re-read a few times to parse it out

django_mongodb_backend/indexes.py
NoahStapp: Can you explain what your initial comment means in this context? You mean disallow `indexing_methods="flat"` as an argument?

blink1073: Yes, I think mixing strings and tuples/lists is a footgun in general

NoahStapp: So if a user wants to have all of their vector fields use `flat` indexing, they'd have to pass `index_methods=["flat", "flat", "flat", ...]` and so on?

blink1073: oof, I see.  yeah, I guess in that case a string makes sense, unfortunately.

blink1073 (approved):
LGTM (other than a lint formatting error)

docs/conf.py
```diff
@@ -56,9 +56,10 @@
 }
 
 # Some pages use JavaScript and anchors won't work right away.
-linkcheck_ignore = [
+linkcheck_anchors_ignore = [
     "https://www.mongodb.com/docs/manual/core/queryable-encryption/install-library/#std-label-qe-csfle-shared-library-download",
     "https://www.mongodb.com/docs/vector-search/index/vector-search-type/#std-label-avs-similarity-functions",
+    "https://www.mongodb.com/docs/vector-search/index/vector-search-type/#about-the-indexing-methods",
 ]
```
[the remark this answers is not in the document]
blink1073: Ah, this is correct

blink1073 (approved):

tests/indexes_/test_search_indexes.py
```diff
@@ -420,3 +450,48 @@ def test_similarities_list(self):
...
             with connection.schema_editor() as editor:
                 editor.remove_index(index=index, model=SearchIndexTestModel)
+
+    def test_indexing_methods_list(self):
+        index = VectorSearchIndex(
+            name="recent_test_idx",
+            fields=["vector_float", "vector_integer"],
+            similarities="cosine",
+            indexing_methods=["flat", "hnsw"],
+        )
```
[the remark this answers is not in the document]
NoahStapp: This should already be covered by the `@skipUnlessDBFeature("supports_search")` annotation on the test class.

docs/ref/models/indexes.rst
```diff
@@ -103,6 +103,18 @@ minutes, depending on the size of the collection.
...
+    provide this value either as a string, in which case that value will be
     applied to all vector fields, or a list or tuple of values with a
     similarity corresponding to each vector field.
+
+    The optional ``indexing_methods`` keyword argument selects the
+    :ref:`indexing method <vector-search:avs-vector-index-method>` for each
+    vector field. Available values are ``"hnsw"`` (the server's default if unspecified)
+    and ``"flat"``. You can provide this value
+    either as a string, in which case that value will be applied to all
+    vector fields, or as a list or tuple of values with an indexing method
```
timgraham, line 115: The line wrapping is inconsistent and some text extends past 79 characters (also the last line of the release note).

docs/ref/models/indexes.rst
```diff
@@ -103,6 +103,18 @@ minutes, depending on the size of the collection.
...
+    corresponding to each vector field (requires MongoDB 8.0+).
```
timgraham, line 116: Is the version requirement correct? According to AI: "The indexingMethod option (used specifically to configure index types like flat instead of the default hnsw graph) was introduced and is supported starting in MongoDB 7.0+ for MongoDB Vector Search." `test-python-atlas.yml` uses MongoDB/Atlas 7 and doesn't skip `test_indexing_methods_list` I believe.

django_mongodb_backend/indexes.py
```diff
@@ -308,8 +308,9 @@ class VectorSearchIndex(SearchIndex):
     suffix = "vsi"
     VALID_FIELD_TYPES = frozenset(("boolean", "date", "number", "objectId", "string", "uuid"))
     VALID_SIMILARITIES = frozenset(("cosine", "dotProduct", "euclidean"))
+    VALID_INDEXING_METHODS = frozenset(("hnsw", "flat"))
```
timgraham, line 311: I would alphabetize this in the list of class attributes.

docs/ref/models/indexes.rst
```diff
@@ -103,6 +103,18 @@ minutes, depending on the size of the collection.
     Available values for the required ``similarities`` keyword argument are
     ``"cosine"``, ``"dotProduct"``, and ``"euclidean"`` (see
     :ref:`vector-search:avs-similarity-functions` for how to choose). You can
-    provide this value either a string, in which case that value will be
+    provide this value either as a string, in which case that value will be
```
timgraham, line 106: I advocate for resisting the urge to make unrelated fixes in a feature or bug fix since it creates a confusing history and eliminates the opportunity to backport the typo fix to older versions of the docs.

tests/indexes_/test_search_indexes.py
```diff
@@ -420,3 +450,48 @@ def test_similarities_list(self):
         finally:
             with connection.schema_editor() as editor:
                 editor.remove_index(index=index, model=SearchIndexTestModel)
+
+    def test_indexing_methods_list(self):
```
timgraham, line 454: It seems a test for the `indexing_methods` as single value applied to multiple fields is missing (see `test_multiple_fields` for the corresponding test for `similarities`).

docs/ref/models/indexes.rst
NoahStapp: Should this be codified into a linting rule?

docs/ref/models/indexes.rst
NoahStapp: I could only find references to 8.0, but if it runs and passes the test then 7.0+ must be correct.

django_mongodb_backend/indexes.py
NoahStapp: Should this also be a linting rule?

docs/ref/models/indexes.rst
NoahStapp: Ah, the distinction is self-hosted search vs Atlas. Any Vector-search enabled Atlas cluster will support `indexingMethods`, but self-hosted requires 8+ since that's when support was added for vector search there. Removing the requirement callout here makes sense in that context.

NoahStapp:
Opened https://github.com/mongodb/django-mongodb-backend/pull/581 to address your comments @timgraham.

django_mongodb_backend/indexes.py
timgraham: I wouldn't call it a hard and fast rule. There could be some logical grouping or order of attributes. Just seeing two VALID_-prefixed variables, it's my intuition to order them. Call it a Tim-nit.

docs/ref/models/indexes.rst
timgraham: Sure, https://github.com/mongodb/django-mongodb-backend/pull/583.

PR #585, opened 2026-06-27
INTPYTHON-1005 Make makemigrations order embedded model creation before dependent operations

timgraham:
INTPYTHON-1005

Embedded model fields reference a model but, unlike relational fields, have no remote_field, so the autodetector didn't make the referencing model's CreateModel (or an AddField/AlterField that adds an embedded field) depend on the embedded model's creation. The embedding operation could be ordered before the embedded model's CreateModel, leaving a dangling reference that can't be resolved while a partial migration state is rendered.

Add a CREATE dependency on each embedded model so it's created first, whether the field is part of a CreateModel, AddField, or AlterField.


blink1073:
There are a couple of merge conflicts

blink1073 (approved):
LGTM!

PR #594, opened 2026-07-02
INTPYTHON-998 Add a table mapping Django and backend branches

timgraham:
(There will be more PRs for this ticket, but this stands alone.)


aclark4life (approved):
Nice thanks

PR #608, opened 2026-08-04
INTPYTHON-1006 Update to Django 6.1

aclark4life:
Re-opens the work from #422 on a fork branch to get fresh CI runs. #422's Actions runs are from 2026-07-02, past GitHub's 30-day re-run window, so they can never be re-queued.

Same head commit as #422: a4787d0a0bdcdff18333a0204135fa2a6283a19d

Refs #422


pyproject.toml
```diff
@@ -21,7 +21,7 @@ keywords = [
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Framework :: Django",
-    "Framework :: Django :: 6.0",
```
NoahStapp, line 24: Don't we need to bump our requirement versions to >= 6.1.0?

django_mongodb_backend/schema.py
```diff
@@ -182,112 +164,77 @@ def remove_field(self, model, field):
...
-            self._remove_model_indexes(
-                field.embedded_model, parent_model=model, column_prefix=new_path
-            )
 
-    def _remove_model_indexes(self, model, column_prefix="", parent_model=None):
-        """
-        When removing an EmbeddedModelField, the indexes need to be removed
-        recursively.
-        """
+    def _remove_model_indexes(self, model):
```
NoahStapp: Is this function used anywhere still?

NoahStapp (changes requested):

pyproject.toml
aclark4life: yes, done in 96717896367575416f011eb23ab6f4c41b887c29

django_mongodb_backend/schema.py
aclark4life: Removed in 96717896367575416f011eb23ab6f4c41b887c29

django_mongodb_backend/gis/features.py
```diff
@@ -26,9 +26,6 @@ def django_test_skips(self):
         skips = super().django_test_skips
         skips.update(
             {
-                "inspectdb not supported.": {
```
NoahStapp, line 29: We support `inspectdb` now?

NoahStapp (changes requested):

django_mongodb_backend/gis/features.py
aclark4life: No, still unsupported. Django 6.1 adds the `DatabaseFeatures.supports_inspectdb` flag (see the 6.1 release notes: "Set the new `DatabaseFeatures.supports_inspectdb` attribute to `False` if the management command isn't supported"), and we set it to `False` in `features.py`. `gis_tests.inspectapp.tests.InspectDbTests` is now decorated with `@skipUnlessDBFeature("supports_inspectdb")` upstream (as are the `inspectdb.tests` cases), so these manual skips are redundant. Done in 297b30c.

aclark4life: No, but [Django 6.1 added a flag `supports_inspectdb`](https://docs.djangoproject.com/en/6.1/releases/6.1/#database-backend-api) and we [set it to False in features.py](https://github.com/mongodb/django-mongodb-backend/commit/297b30cabc907513719c272551ea242da6788d81), so it's redundant here.

NoahStapp (approved):

PR #610, opened 2026-08-04
INTPYTHON-994 Make QuerySet.exclude() omit documents with absent JSONField keys

aclark4life:
[INTPYTHON-994](https://jira.mongodb.org/browse/INTPYTHON-994)

Re-opens the work from #576 on a fork branch. Original author: @timgraham. Same commits as #576 (2516d1e), unmodified.

`exclude(value__key=X)` and negated `Q` objects on `JSONField` key lookups wrongly returned documents that don't contain the key. SQL databases treat the comparison on a missing key as `NULL` (three-valued logic), so `NOT(...)` excludes those rows; MongoDB's `$nor`/`$not` instead treat a missing key as "not equal" and included them.

## Changes in this PR

`django_mongodb_backend/fields/json.py`

- Adds `key_transform_negated()`, a private helper that compiles `NOT(<JSONField key lookup>)` as *key exists* AND the negated positive MQL, matching SQL three-valued semantics. It handles both the `$expr` and path-based (`can_use_path`) compilation modes, walking back to the root column so the null check applies to the whole field.
- Registers it as `as_mql_negated` on `KeyTransformExact`, `KeyTransformIn`, `KeyTransformNumericLookupMixin`, and `KeyTransformTextLookupMixin`, so every comparison lookup on a JSON key gets the corrected negation.

`django_mongodb_backend/query.py`

- Adds `_has_custom_negation()` and `_negate()`, private helpers that push a negation down to the leaves via De Morgan's laws.
- `where_node()` now takes that path when a negated `AND`/`OR` node contains a lookup defining `as_mql_negated`. Each such lookup supplies its own SQL-correct negation; every other lookup keeps its existing `$nor`/`$not` compilation, and `EmptyResultSet`/`FullResultSet` are swapped and propagated as the De Morgan transformation requires.

`as_mql_negated` is a new private extension point on lookups — an opt-in hook, so lookups that don't define it are unaffected.

Docs

- Adds a 6.0.x release note for the bug fix.
- Removes the corresponding entry from `docs/topics/known-issues.rst`, leaving the unrelated JSON `null` / SQL `null` caveat in place.

## Test Plan

No tests are added in this repo. JSONField key-lookup coverage lives in the Django test suite that CI runs against the [mongodb-forks/django](https://github.com/mongodb-forks/django) branch (`model_fields.test_jsonfield`)

### Checklist for Author

- [x] Did you update the changelog (if necessary)?
- [ ] Is the intention of the code captured in relevant tests? — covered by the upstream Django suite, not by tests in this repo (see Test Plan)
- [ ] If there are new TODOs, has a related JIRA ticket been created? — no new TODOs

### Checklist for Reviewer


blink1073:
We need to remove the test skips as part of this PR.

